Issue #124, opened 2020-09-17
Separate particle filter from the model 


giordano:
As it is now, this package blends the particle filter part with a single hard-coded model.  We want to disentangle this and separate the particle filter from user-provided models.  In this issues we'll collect the steps to do this

# Model

## Functions
* Initialise
  * Arrays
  * Initial state
  * Stations
  * Random field
* Update
  * Time step
  * Add noise
* Get observations
  * Add noise

Model functions will all return (or modify in place) a data structure that contains all of the structures below.

## Data Structures

* States
* Observations
* Stations
* Internal model arrays
* Random field
* Buffer(s)

## Parameters

    nx::Int
    ny::Int
    x_length::T
    y_length::T
    dx::T
    dy::T

    n_state_var::Int

    station_filename::String
    nobs::Int
    station_distance_x::T
    station_distance_y::T
    station_boundary_x::T
    station_boundary_y::T
    source_size::T
    bathymetry_setup::T

    lambda::T
    nu::T
    sigma::T

    padding::Int
    primes::Bool
    obs_noise_std::T

    lambda_initial_state::T
    nu_initial_state::T
    sigma_initial_state::T
    particle_initial_state::String

    absorber_thickness_fraction::T
    boundary_damping::T
    cutoff_depth::T

# Filter

Filter expects two models as arguments, one for the true state and observations, and one for the particle state and observations. If only one model is passed, it is used for both.

## Functions

* Initialise
* Get weights
* Resample
* Statistics
* IO
* Diagnostics

## Data Structures

* Statistics
* Weights
* Resampling indices
* Buffer(s)

## Parameters

    n_time_step::Int
    n_integration_step::Int
    time_step::T
    verbose::Bool
    nprt::Int
    random_seed::Int
    enable_timers::Bool
    master_rank::Int

    output_filename::String
    state_prefix::String
    title_avg::String
    title_var::String
    title_syn::String
    title_grid::String
    title_stations::String
    title_params::String

PR #128, opened 2020-09-28
Separate model from particle filter

giordano:
Still a work in progress, but not too far from the goal of #124.  I believe that in the end this package should only provide the filter, the model we have now will become an implementation of a model, to be used in the tests.

## Design

The model needs to define a bunch of things:

* a custom data structure, called here `ModelData`, but the name isn't relevant.  The particle filter doesn't need to know _anything_ about it, but it'll call some functions on it
* the init function, which should be passed as argument to `run_particle_filter`.  This function will return an instance of `ModelData`
* the method `update_truth!(model_data::ModelData, params)`, which returns the observations of the truth
* the method `update_particles!(model_data::ModelData, nprt_per_rank, params)`, which returns the observations of the particles
* the method `get_particles(::ModelData)` which returns the particles.  I wish this wasn't necessary, but I don't see many ways around this

In addition, right now I use a couple of more methods, which I feel like are unnecessary:

* ~`get_buffer(::ModelData)`: we may just allocate a different buffer in the particle filter, if necessary~  This buffer was used only in the filter, so I simply moved it there and removed from the model
* `get_truth(::ModelData)`: it's used only in the return value of `run_particle_filter`, we may just return the whole `ModelData`, the user will decide what to do with it.

## TODO

Things that are still entangled:

* [x] ~input parameters: instead of creating complicated custom data types for the parameters of the particle filter _and_ of the model, we may have a single dictionary-like data type.  It isn't super fancy, but should do the job good enough and could be flexible enough for multiple different model.~
* [x] ~writing to the output HDF5 is strongly coupled between filter and model.  I'm thinking about having the model to define a function `write_snapshot(::ModelData)`, and maybe also `write_initial_state(::ModelData)`, but as it is now the I/O functions write at the same time all data, in a somewhat correlated way.  We need to better figure out how to efficiently do this, especially to avoid opening/closing the HDF5 multiple times.~ I simply moved most of I/O to the model.  This isn't great because now the model code is quite large, much larger than the filter itself, but I/O is really mostly specific to the model.
* [x] As far as I can see, the particle filter doesn't use `rng`, I think it can be set directly by the model.


src/ParticleDA.jl
```diff
@@ -5,81 +5,13 @@ import Future
...
-include("llw2d.jl")
+include("model.jl")
 include("io.jl")
 
 using .Default_params
-using .LLW2d
-
-# grid-to-grid distance
-get_distance(i0, j0, i1, j1, dx, dy) =
-    sqrt((float(i0 - i1) * dx) ^ 2 + (float(j0 - j1) * dy) ^ 2)
```
giordano, line 19: This function doesn't seem to be used anywhere, apart from the tests

src/ParticleDA.jl
```diff
@@ -516,87 +202,51 @@ function tdac(params::Parameters, rng::AbstractVector{<:Random.AbstractRNG})
...
-        resampling_indices = Vector{Int}(undef,params.nprt)
+    resampling_indices = Vector{Int}(undef, params.nprt)
 
-        avg_arr = Array{Float64,3}(undef, params.nx, params.ny, params.n_state_var)
-        var_arr = Array{Float64,3}(undef, params.nx, params.ny, params.n_state_var)
+    # TODO: these variables should be set in a better way
+    nx, ny, n_state_var = params.nx, params.ny, params.n_state_var
+    statistics = Array{SummaryStat{T}, 3}(undef, nx, ny, n_state_var)
+    avg_arr = Array{T,3}(undef, nx, ny, n_state_var)
+    var_arr = Array{T,3}(undef, nx, ny, n_state_var)
```
giordano, line 225: I think the sizes of these arrays should come from the model.  It needs to somehow pass the size `(nx, ny, n_state_var)` to the filter.

src/ParticleDA.jl
```diff
@@ -516,87 +202,51 @@ function tdac(params::Parameters, rng::AbstractVector{<:Random.AbstractRNG})
...
     if(params.verbose && my_rank == params.master_rank)
-
-        @timeit_debug timer "IO" begin
-            write_grid(params)
-            write_params(params)
-            write_stations(stations.ist, stations.jst, params)
-            unpack_statistics!(avg_arr, var_arr, statistics)
-            write_snapshot(states.truth, avg_arr, var_arr, weights, 0, params)
-        end
+        # @timeit_debug timer "IO" write_initial_state(model_data, params)
```
giordano, line 233: I/O temporarily disabled, until we figure out how to do this properly

src/ParticleDA.jl
```diff
@@ -623,15 +273,15 @@ function tdac(params::Parameters, rng::AbstractVector{<:Random.AbstractRNG})
...
-        @timeit_debug timer "Mean and Var" get_mean_and_var!(statistics, states.particles, params.master_rank)
+        @timeit_debug timer "Mean and Var" get_mean_and_var!(statistics, get_particles(model_data), params.master_rank)
 
         if my_rank == params.master_rank && params.verbose
 
             @timeit_debug timer "IO" begin
-                unpack_statistics!(avg_arr, var_arr, statistics)
-                write_snapshot(states.truth, avg_arr, var_arr, weights, it, params)
+                # unpack_statistics!(avg_arr, var_arr, statistics)
+                # write_snapshot(states.truth, avg_arr, var_arr, weights, it, params)
```
giordano, line 284: I/O temporarily disabled, until we figure out how to do this properly

src/ParticleDA.jl
```diff
@@ -662,14 +312,14 @@ function tdac(params::Parameters, rng::AbstractVector{<:Random.AbstractRNG})
             timer_chars = MPI.Gather(chr_timer, params.master_rank, MPI.COMM_WORLD)
 
             if my_rank == params.master_rank
-                write_timers(length_timer, my_size, timer_chars, params)
+                # write_timers(length_timer, my_size, timer_chars, params)
```
giordano, line 315: I/O temporarily disabled, until we figure out how to do this properly

src/ParticleDA.jl
```diff
@@ -723,11 +373,11 @@ function tdac(path_to_input_file::String, rng::AbstractRNG)
         [m; accumulate(Future.randjump, fill(big(10)^20, nthreads()-1), init=m)]
     end;
 
-    return tdac(params, rng_vec)
+    return run_particle_filter(params, rng_vec, init)
 
 end
 
-function tdac(path_to_input_file::String = "")
+function run_particle_filter(path_to_input_file::String = "")
```
giordano, line 380: Also this function should get the `init` function as argument

src/ParticleDA.jl
```diff
@@ -746,11 +396,11 @@ function tdac(path_to_input_file::String = "")
 
     params = MPI.bcast(params, 0, MPI.COMM_WORLD)
 
-    return tdac(params)
+    return run_particle_filter(params)
 
 end
 
-function tdac(params::Parameters)
+function run_particle_filter(params::Parameters)
```
giordano, line 403: Also this function should get the `init` function as argument

src/model.jl
```diff
@@ -0,0 +1,379 @@
...
+
+struct ModelData
+    states
+    observations
+    stations
+    field_buffer
+    background_grf
+    model_matrices
+    rng
+end
```
giordano, line 311: Once we finalise the data types, this needs to be concretely annotated, maybe using parameters, for better performance

test/runtests.jl
```diff
@@ -62,9 +62,6 @@ end
 @testset "ParticleDA unit tests" begin
     dx = dy = 2e3
 
-    @test ParticleDA.get_distance(3/2000, 4/2000, 0, 0, dx, dy) == 5
-    @test ParticleDA.get_distance(10, 23, 5, 11, dx, dy) == 26000.0
```
giordano, line 66: As above, this function doesn't seem to be used anywhere

src/model.jl
giordano: Done.  Much to my surprise, this doesn't give a large performance improvement.

src/ParticleDA.jl
```diff
@@ -434,20 +120,14 @@ function get_mean_and_var!(statistics::Array{SummaryStat,3},
...
         avg[idx] = statistics[idx].avg
         var[idx] = statistics[idx].var
     end
 end
 
-function copy_states!(states::StateVectors, resampling_indices::Vector{Int}, my_rank::Int, nprt_per_rank::Int)
-
-    copy_states!(states.particles, states.buffer, resampling_indices, my_rank, nprt_per_rank)
-
-end
```
giordano, line 449: This method isn't needed anymore, we're calling directly the one below.

src/ParticleDA.jl
```diff
@@ -616,91 +266,91 @@ function tdac(params::Parameters, rng::AbstractVector{<:Random.AbstractRNG})
...
         @warn "Input file " * path_to_input_file * " not found, using default parameters"
-        params = get_params()
+        user_input_dict = Dict()
     end
-    return params
+    return user_input_dict
 
 end
 
-function tdac(path_to_input_file::String, rng::AbstractRNG)
```
giordano, line 703: The diff is screwing things up a bit, what I did here was to remove this method and keep only the two other methods below

tkoskela (commented):
Overall I think this looks pretty good! I added a few comments.

test/model/model.jl
```diff
@@ -0,0 +1,613 @@
...
+    observations::C
+    stations::D
+    field_buffer::E
+    background_grf::F
+    model_matrices::G
+    rng::H
+end
+ParticleDA.get_particles(d::ModelData) = d.states.particles
+# TODO: we should probably get rid of `get_truth`: it is only used as return
+# value of `particle_filter`, we may just return the whole `model_data`.
```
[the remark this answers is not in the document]
giordano: Yes, no problem having the same signature for `update_truth!` and `update_particles!`, it makes a lot of sense.

I'm going to pass `nprt_per_rank` also to `update_truth!`, it'd simply be ignored in the implementation of the model in the tests.  If the user wants to pass a single model they can do:
```julia
function ParticleDA.update_particles!(data::ModelData, nprt_per_rank)
    # ...
end

function ParticleDA.update_truth!(data::ModelData, nprt_per_rank)
    ParticleDA.update_particles!(data::ModelData, nprt_per_rank)
end
```

test/model/model.jl
```diff
@@ -0,0 +1,613 @@
+module LLW2d
+
+using ParticleDA
+
+using Random, Distributions, Base.Threads, GaussianRandomFields, HDF5
+using ParticleDA.Default_params
+using DelimitedFiles
+import Future
+
+include("llw2d.jl")
```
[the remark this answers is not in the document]
giordano: Done: now the file `llw2d.jl` is completely unchanged, the file `model.jl` defines the `Model` module.

src/ParticleDA.jl
```diff
@@ -422,9 +113,9 @@ function stats_reduction(S1::SummaryStat, S2::SummaryStat)
 
 end
 
-function get_mean_and_var!(statistics::Array{SummaryStat,3},
-                                    particles::AbstractArray{T,4},
-                                    master_rank::Int) where T
+function get_mean_and_var!(statistics::Array{SummaryStat{T},3},
+                           particles::AbstractArray{T,4},
```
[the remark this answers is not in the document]
giordano: BTW, this is basically the same comment as https://github.com/Team-RADDISH/ParticleDA.jl/pull/128#discussion_r496034922 :sweat_smile:  Maybe the size could be passed as a tuple by the `init` function?

tkoskela (changes requested):
Can you please update the usage section in README.md?

giordano:
I've reduced the README to the very minimum, when this PR will be merged, the documentation will be published to a website based on GitHub Pages.

tkoskela (approved):

PR #133, opened 2020-12-01
Implement optimal particle filter from Alex

tkoskela:
Implementation of [Alex Optimal Particle Filter code](https://github.com/Team-RADDISH/mini_apps/blob/master/optimal_particle_filter/Sample_Optimal_Height_Proposal.jl)

Includes validation and benchmark (although with hard-coded file path)

```
old mean height:   3.534 ms (2450 allocations: 4.38 MiB)
new mean height:   541.213 μs (425 allocations: 446.55 KiB)
old sampling:   8.128 ms (157036 allocations: 11.14 MiB)
new sampling:   698.481 μs (552 allocations: 676.05 KiB)
```


src/OptimalFilter.jl
```diff
@@ -0,0 +1,504 @@
...
+    first_column_covariance_extended_grid!(matrices.rho_bar, params)
+    covariance_stations_grid!(matrices.R12, params, stations)
+    covariance_stations_extended_grid!(matrices.R21_bar, params, stations)
+    covariance_stations!(matrices.R22, params, stations)
+    matrices.R22_inv .= inv(matrices.R22)
+    matrices.R12_invR22 .= matrices.R12 * matrices.R22_inv
+
+    fourier_coeffs = Vector{ComplexF64}(undef, n1_bar)
+    normalized_inverse_2d_fft!(fourier_coeffs, matrices.rho_bar, params)
+    matrices.Lambda .= sqrt(4 * params.nx * params.ny) * real.(fourier_coeffs)
```
giordano, line 262: This is completely in-place, it doesn't create the temporary array for the multiplication on the right-hand side
```suggestion
    matrices.Lambda .= sqrt(4 * params.nx * params.ny) .* real.(fourier_coeffs)
```

src/OptimalFilter.jl
```diff
@@ -0,0 +1,504 @@
...
+
+    WHbar_R12 = Matrix{C}(undef, n1_bar, params.nobs)
+    for i in 1:params.nobs
+        normalized_2d_fft!(@view(WHbar_R12[:,i]), @view(matrices.R21_bar[i,:]), params)
+    end
+    KH = Diagonal(matrices.Lambda)^(-1/2)*WHbar_R12
+    matrices.K .= KH'
+
+    A = real.(matrices.R22 - matrices.K*KH)
```
giordano, line 271: Some more dots, for good measure
```suggestion
    A = real.(matrices.R22 .- matrices.K.*KH)
```

src/OptimalFilter.jl
```diff
@@ -0,0 +1,504 @@
...
+    if ishermitian(A)
+        matrices.L .= cholesky(A).L
+    end
+
+    matrices.mu20 .= params.sigma_obs^(-2) * (matrices.R22 - I(params.nobs) * params.sigma_obs^2)
```
giordano, line 276: (almost) completely in-place:
```suggestion
    matrices.mu20 .= params.sigma_obs^(-2) .* (matrices.R22 .- I(params.nobs) .* params.sigma_obs^2)
```

src/OptimalFilter.jl
```diff
@@ -0,0 +1,504 @@
...
+    # This corresponds to mu2 in Alex's code.
+    # TODO: Check if WΛWH is linear and you could swap the order
+    offline_matrices.buf2 .-= offline_matrices.buf1
+
+    mu10 = Vector{T}(undef, params.nobs)
+
+    # Loop over particles
+    for iprt = 1:params.nprt
+
+        mu10 .= offline_matrices.R22_inv * get_values_at_stations(@view(height[iprt,:,:]), stations)
```
giordano, line 329: You can use `mul!` if you want to do the multiplication in-place

src/OptimalFilter.jl
```diff
@@ -0,0 +1,504 @@
...
+        mu10_sparse = sparse(stations.ist, stations.jst, mu10, params.nx+1, params.ny+1)
+
+        # Compute decomposition of height values at stations times the inverse covariance matrix
+        # The argument corresponds to mu10 and the outcome to mu11 in Alex's code
+        WΛWH_decomposition!(offline_matrices.buf1, mu10_sparse, offline_matrices, params)
+
+        # Compute the mean for the ith particle using mu2 and mu11
+        # Skip storing the temporary mu1 in Alex's code
+        mean[iprt,:,:] .= @view(height[iprt,:,:]) - offline_matrices.buf1 + offline_matrices.buf2
```
giordano, line 338:
```suggestion
        mean[iprt,:,:] .= @view(height[iprt,:,:]) .- offline_matrices.buf1 .+ offline_matrices.buf2
```

giordano (commented):
Some stylistic comments, mostly dots :smile:

src/OptimalFilter.jl
```diff
@@ -0,0 +1,504 @@
...
+        x_ext = x
+    elseif (params.x_length <= x) & (x <= 2 * params.x_length)
+        x_ext = 2 * params.x_length - x
+    else
+        @error "value of x is out of bounds"
+    end
+
+    if (0 <= y) & (y <= params.y_length)
+        y_ext = y
+    elseif (params.y_length <= y) & (y <= 2 * params.y_length)
```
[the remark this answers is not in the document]
tkoskela: Oh, nice, that looks much neater

src/OptimalFilter.jl
tkoskela: Oops, K*KH is a real matrix product

giordano: Ugh :disappointed: :facepalm:

src/OptimalFilter.jl
```diff
@@ -0,0 +1,512 @@
...
+    calculate_mean_height!(mean, height, offline_matrices, observations, stations, params)
+
+    #TODO: Can you create a Normal(0,1) distribution of complex type?
+    nd = Normal(0,1)
+
+    i_n1 = LinearIndices((params.nx+1, params.ny+1))
+    i_n1_bar = LinearIndices((2*params.nx, 2*params.ny))
+
+    for iprt in 1:2:params.nprt
+        # TODO randn(ComplexF64) seems to set variance = 1/2. Could not find how to change that.
```
giordano, line 366: I'm not sure what you mean:
```julia
julia> using Statistics

julia> M = randn(ComplexF64, (100, 100));

julia> std(M)
1.0015637176495376

julia> mean(M)
-0.005198967157672335 - 0.0066281896290166595im
```
The mean is about 0 and the standard deviation is about 1, what's the issue?

tkoskela: Hmm, ok, I didn't actually test that. I trusted what it said here https://docs.julialang.org/en/v1/stdlib/Random/#Base.randn

Is the difference whether the variance of the complex number is 1 or the variances of the real and imaginary parts are 1? In Alex's code it was the latter.

giordano: The documentation is correct.  This is how complex `randn` is implemented: https://github.com/JuliaLang/julia/blob/a7848a28e5d450dcc48ef99320d3220618dba501/stdlib/Random/src/normal.jl#L86-L91.  Having the real and the imaginary parts with standard deviation 1, individually, means that the complex numbers have standard deviation `sqrt(2)`:
```julia
julia> using Statistics

julia> M = complex.(randn(100, 100), randn(100, 100));

julia> std(M)
1.4117042745940547

julia> mean(M)
-0.006128621492362236 - 0.011696050838561518im
```
Looking at the paper, I'm mildly confident we want to have complex numbers with standard deviation 1, not `sqrt(2)`

Project.toml
```diff
@@ -5,10 +5,14 @@ version = "0.1.0"
...
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
giordano, line 15: We don't need `Test` here, only in the environment of the tests
```suggestion
```

Also, please add a compat entry for `FFTW` below and in the environment of the tests:
```toml
FFTW = "1"
```
should be fine

src/ParticleDA.jl
```diff
@@ -27,6 +36,16 @@ by the user with the above signature, specifying the type of `model_data`.
...
 
+"""
+    ParticleDA.get_stations(model_data) -> station_coordinates
+
+Return an array of coordinates of the points of observation. This method is intended
+to be extended by the user with the above signature, specifying the type of `model_data`.
+Required for optimal filter only.
+"""
+function get_stations end
+function get_rng end
```
giordano, line 47: Missing docstring?

src/OptimalFilter.jl
```diff
@@ -0,0 +1,382 @@
...
+                                 offline_matrices::OfflineMatrices, online_matrices::OnlineMatrices,
+                                 observations::AbstractVector{T}, stations, model_params, filter_params,
+                                 rng::Random.AbstractRNG) where T
+
+    @assert filter_params.nprt % 2 == 0 "Number of particles must be even"
+
+    calculate_mean_height!(online_matrices.mean, height, offline_matrices, observations, stations, model_params, filter_params)
+
+    #TODO: Can you create a Normal(0,1) distribution of complex type?
+    nd = Normal(0,1)
```
giordano, line 334: In any case we don't really need to define this, `randn` has the same properties

src/OptimalFilter.jl
```diff
@@ -0,0 +1,382 @@
...
+
+    i_n1 = LinearIndices((model_params.nx+1, model_params.ny+1))
+    i_n1_bar = LinearIndices((2*model_params.nx, 2*model_params.ny))
+
+    for iprt in 1:2:filter_params.nprt
+        # TODO randn(ComplexF64) seems to set variance = 1/2. Could not find how to change that.
+        # TODO we could pre-create all our random numbers in one go before the loop, would that be faster?
+        e1 = rand(rng, nd, 4*model_params.nx*model_params.ny) .+ rand(rng, nd, 4*model_params.nx*model_params.ny)im
+        e2 = rand(rng, nd, model_params.nobs) .+ rand(rng, nd, model_params.nobs)im
```
giordano, line 343: This doesn't need a special distribution:
```suggestion
        e1 = complex.(randn(rng, 4*model_params.nx*model_params.ny), randn(rng,4*model_params.nx*model_params.ny))
        e2 = complex.(randn(rng, model_params.nobs), randn(rng, model_params.nobs))
```

src/OptimalFilter.jl
```diff
@@ -0,0 +1,382 @@
...
+
+    for iprt in 1:nprt
+
+        log_weights[iprt] = -0.5 * (obs - obs_model[:,iprt])' * matrices.R22_inv * (obs - obs_model[:,iprt])
+
+    end
+
+    #Normalization is done in a separate function due to parallelism optimisation
+
+end
```
giordano, line 382:
```suggestion
end

```

src/OptimalFilter.jl
tkoskela: This change breaks the validation test. Any idea why? I'm guessing the random stream gets screwed up somehow but I don't see how

giordano: It's not unlikely that the stream of numbers generated by `rand(Normal(1, 0), ...)` is different from `randn(...)`, but their statistical properties are the same

giordano: Actually, the stream seems to be the same:
```julia
julia> using Random, Distributions

julia> rng = Random.seed!(42)
MersenneTwister(42)

julia> randn(rng, 10)
10-element Vector{Float64}:
 -0.5560268761463861
 -0.444383357109696
  0.027155338009193845
 -0.29948409035891055
  1.7778610980573246
 -1.14490153172882
 -0.46860588216767457
  0.15614346264074028
 -2.641991008076796
  1.0033099014594844

julia> rng = Random.seed!(42)
MersenneTwister(42)

julia> rand(rng, Normal(0, 1), 10)
10-element Vector{Float64}:
 -0.5560268761463861
 -0.444383357109696
  0.027155338009193845
 -0.29948409035891055
  1.7778610980573246
 -1.14490153172882
 -0.46860588216767457
  0.15614346264074028
 -2.641991008076796
  1.0033099014594844
```
:thinking:

giordano: This is how sampling of `Normal(μ, σ)` is implemented: https://github.com/JuliaStats/Distributions.jl/blob/863844c88e4153af13996f571fcc612d159de542/src/univariate/continuous/normal.jl#L243, so yeah, there is no reason why it should be different :confused:

tkoskela: Yeah, I checked the same. Oddly enough in the validation test it does not work.

If I change both sides of the test to use `randn` then it works again 🤔

test/runtests.jl
```diff
@@ -256,3 +256,103 @@ end
...
+    model_params = ParticleDA.get_params(ModelParameters, params_dict["model"]["llw2d"])
+
+    stations = (ist = st.st_ij[:,1].+1, jst = st.st_ij[:,2].+1)
+
+    h(x,y) = sin(x)^2 + cos(y)^2
+    height = zeros(model_params.nx, model_params.ny, filter_params.nprt)
+    x = (1:model_params.nx) .* 2 * pi / model_params.nx
+    y = (1:model_params.ny) .* 4 * pi / model_params.ny
+    for i = 1:filter_params.nprt
+        height[:,:,i] = h.(x',y)
```
giordano, line 327: This becomes in-place :wink: 
```suggestion
        height[:,:,i] .= h.(x',y)
```

src/OptimalFilter.jl
```diff
@@ -0,0 +1,396 @@
...
+
+# Extended covariance function /bar r(x,y), equation 8 of Dietrich and Newsam 96
+function extended_covariance(x::T, y::T, model_params, filter_params) where T
+
+    if 0 <= x <= model_params.x_length
+        x_ext = x
+    elseif model_params.x_length <= x <= 2 * model_params.x_length
+        x_ext = 2 * model_params.x_length - x
+    else
+        @error "value of x is out of bounds"
```
giordano, line 55: Do you want to thrown an error here?  `@error` is a macro for logging

src/OptimalFilter.jl
```diff
@@ -0,0 +1,396 @@
...
+end
+
+# Allocate and compute matrices that do not depend on time-dependent variables (height and observations).
+function init_offline_matrices(model_params, filter_params, stations)
+
+    n1 = model_params.nx * model_params.ny # number of elements in original grid
+    n1_bar = model_params.nx_ext * model_params.ny_ext     # number of elements in extended grid
+
+    F = Float64
+    C = ComplexF64
```
giordano, line 229: If you want to infer `C` from `F`, you can do
```suggestion
    C = complex(F)
```
for when we can use different float types

src/OptimalFilter.jl
```diff
@@ -0,0 +1,396 @@
...
+                               )
+
+    first_column_covariance_extended_grid!(matrices.rho_bar, model_params, filter_params)
+    covariance_stations_grid!(matrices.R12, model_params, filter_params, stations)
+    covariance_stations_extended_grid!(matrices.R21_bar, model_params, filter_params, stations)
+    covariance_stations!(matrices.R22, model_params, filter_params, stations)
+    matrices.R22_inv .= inv(matrices.R22)
+    matrices.R12_invR22 .= matrices.R12 * matrices.R22_inv
+
+    fourier_coeffs = Vector{ComplexF64}(undef, n1_bar)
```
giordano, line 253:
```suggestion
    fourier_coeffs = Vector{C}(undef, n1_bar)
```

src/OptimalFilter.jl
```diff
@@ -0,0 +1,396 @@
...
+end
+
+# Allocate memory for matrices that will be updated during the time stepping loop.
+function init_online_matrices(model_params, filter_params)
+
+    n1 = model_params.nx * model_params.ny # number of elements in original grid
+    n1_bar = model_params.nx_ext * model_params.ny_ext # number of elements in extended grid
+
+    C = ComplexF64
+    F = Float64
```
giordano, line 283: Likewise
```suggestion
    F = Float64
    C = complex(F)
```

src/OptimalFilter.jl
```diff
@@ -0,0 +1,396 @@
...
+    end
+
+end
+
+function sample_height_proposal!(height::AbstractArray{T,3},
+                                 offline_matrices::OfflineMatrices, online_matrices::OnlineMatrices,
+                                 observations::AbstractVector{T}, stations::NamedTuple, model_params, filter_params,
+                                 rng::Random.AbstractRNG) where T
+
+    @assert filter_params.nprt % 2 == 0 "Number of particles must be even"
```
giordano, line 347:
```suggestion
    @assert iseven(filter_params.nprt) "Number of particles must be even"
```

test/optimal_filter_validation.jl
```diff
@@ -0,0 +1,897 @@
...
+
+#### Specification of 'Extented' Covariance Function
+#### following Method/Notation in Dietrich & Newsam (96)
+function bar_R(x::Float64, y::Float64, th, gr)
+
+  P = gr.P; Q = gr.Q
+
+  bar_Cov = 0.0
+
+  if ( (0<=x) & (x<=P) & (0<=y) & (y<=Q) )
```
giordano, line 118:
```suggestion
  if 0 <= x <= P && 0 <= y <= Q
```

test/optimal_filter_validation.jl
```diff
@@ -0,0 +1,897 @@
...
+
+    bar_Cov = R(x, y, th)
+
+  elseif ( (P<=x) & (x<=2*P) & (0<=y) & (y<=Q) )
```
giordano, line 122:
```suggestion
  elseif P <= x <= 2*P && 0 <= y <= Q
```

test/optimal_filter_validation.jl
```diff
@@ -0,0 +1,897 @@
...
+
+    bar_Cov = R(2*P-x, y, th)
+
+  elseif ( (0<=x) & (x<=P) & (Q<=y) & (y<=2*Q) )
```
giordano, line 126:
```suggestion
  elseif 0 <= x <= P && Q <= y <= 2*Q
```

test/optimal_filter_validation.jl
```diff
@@ -0,0 +1,897 @@
...
+
+    bar_Cov = R(x, 2*Q-y, th)
+
+  elseif ( (P<=x) & (x<=2*P) & (Q<=y) & (y<=2*Q) )
```
giordano, line 130:
```suggestion
  elseif P <= x <= 2*P && Q <= y <=2*Q
```

test/optimal_filter_validation.jl
tkoskela: I'm just going to leave this file as it is since it's Alex's code. I'm sure you'd have a lot of comments :)

tkoskela:
- [ ] Try using @threads on lines 304 and 333

test/model/model.jl
```diff
@@ -61,8 +61,8 @@ Base.@kwdef struct ModelParameters{T<:AbstractFloat}
     ny::Int = 200
     x_length::T = 400.0e3
     y_length::T = 400.0e3
-    dx::T = x_length / nx
-    dy::T = y_length / ny
+    dx::T = x_length / (nx - 1)
+    dy::T = y_length / (ny - 1)
```
giordano, line 65: Can you remind me why this change?

tkoskela: It comes down to whether we define `nx` and `ny` as being the number of cells or nodes in the grid. We had kind of implicitly defined them as the number of cells, but I think it made more sense to adapt Alex's (and Dietrich&Newsan's) definition of it being the number of nodes. Then `dx`, being the width of a cell is the total width divided by the number of cells, which is 1 less than the number of nodes.

giordano: Ok, thanks!

src/ParticleDA.jl
```diff
@@ -3,19 +3,22 @@ module ParticleDA
...
 # Functions to extend in the model
 
 """
-    ParticleDA.get_grid_size(model_data) -> NTuple{N, Int} where N
+    ParticleDA.get_grid_size(model_data) ->
+        NamedTuple({:nx, :ny, :dx, :dy, :x_length, :y_length},Tuple{Int, Int, Float, Float, Float, Float})
 
-Return a tuple with the size of the data.
+Return a NamedTuple with the parameters that define the size of the grid.
+TODO: It could be better to define a struct for these
```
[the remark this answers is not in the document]
tkoskela: I think functions with different names might be a cleaner solution here. Ie, function `f(d) -> NTuple{N,Int}` that returns (nx,ny,...) and `g(d) -> NTuple{N,Float}` that returns (x_length, y_length, ...). dx and dy can be derived from these so a third function is not really needed (although it might help avoid errors). I'm trying to choose names would be most descriptive. Does this sound right to you:

- `get_grid_dims`: get nx,ny,...
- `get_grid_size`: get x_length, y_length,...
- `get_grid_cell_size`: get dx, dy,...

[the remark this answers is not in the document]
tkoskela: See #157

src/OptimalFilter.jl
```diff
@@ -0,0 +1,406 @@
...
+        normalized_2d_fft!(@view(WHbar_R12[:,i]), @view(matrices.R21_bar[i,:]), fft_plan, fft_plan!, grid_ext)
+    end
+    KH = Diagonal(matrices.Lambda)^(-1/2)*WHbar_R12
+    matrices.K .= KH'
+
+    A = real.(matrices.R22 .- matrices.K*KH)
+
+    if ishermitian(A)
+        matrices.L .= cholesky(A).L
+    end
```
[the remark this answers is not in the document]
tkoskela: Maybe we should do something like
```julia
    if A ≈ Hermitian(A)
        matrices.L .= cholesky(Hermitian(A)).L
    else
        error("R22 - K*KH is not hermitian! Increasing the grid size may help. See Dietrich and Newsam '96 for details.")
    end
```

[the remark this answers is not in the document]
AlexandrosBeskos: Tuomas mentioned in his email: "My suggestion is forcing the matrix to be hermitian when it is already approximately so, and reporting an error when it is not. Does this make sense to you, or do you have other suggestions?" 

I agree, give that the issue is indeed the one that you mention. But I have some comments about the "issue" itself. I do not think the issue is the particular matrix being Hermitian - I think it should be by definition (if I remember well from the Theory). If that recollection is correct, then if you get such a message "of not being Hermitian" as an error message then it *has* to be due to round-off errors. Continuing with my recollection, some issue that indeed arises with this Circulant Embedding method that the algorithm uses is that the matrix you are referring to can have negative eigenvalues - so the Cholesky decomposition method will give an error message. There is no easy way around this. What I mean, is that if one fixes the 2D domain, and the grid, then the only possible solution is some sort of "enlargement" of the Embedding, i.e. we are talking about adapting the core algorithm, so let's not go down this direction. From my understanding, the larger the 2D domain, the more likely that the matrix is positive definite, so in practice we just need to adjust our 2D domain - and grid - in a way that the matrix *is* positive definite, that is the practical solution.

Something I do not understand, is why should the above be related to getting different results at different executions of the algorithm under the same parameters. I mean, once the parameters are chosen, then either there is some issue with this matrix or there is no issue!!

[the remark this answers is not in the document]
giordano: > Something I do not understand, is why should the above be related to getting different results at different executions of the algorithm under the same parameters.

Because if we don't explicitly set `matrices.L` here we access uninitialised memory.  In principle we could always set `matrices.L` to zero before here, but that's slightly slower than using an uninitialised array, which actually made debugging the issue (`matrices.L` not set correctly under certain conditions) much easier.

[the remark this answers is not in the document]
AlexandrosBeskos: Hmmm, I see - more or less. 
Is it something that is easy/difficult to fix?

tkoskela:
How do you feel about merging this now?

giordano:
Let me finish playing with #162 :slightly_smiling_face:

giordano (approved):

PR #142, opened 2021-01-13
Move filter initialisations into a function and group arrays into a struct.

tkoskela:


giordano (approved):
Looks like a good idea!

PR #143, opened 2021-01-14
Update documentation to mention new argument to `run_particle_filter`

giordano:
I forgot to update the documentation in #140.  Also, on a second thought I
believe that it's better to use an instance of the type for dispatch, rather
than the type itself.


tkoskela (approved):
Well done for remembering to update the doc!

Issue #136, opened 2021-01-08
Optimal particle filter


tkoskela:
The [optimal particle filter code](https://github.com/Team-RADDISH/mini_apps/tree/master/optimal_particle_filter) by @AlexandrosBeskos needs to be refactored and integrated into ParticleDA.jl.

- [x] Refactor and optimize the code
  - The refactoring is nearly done in PR #133 
  - Performance has improved due to less memory allocations (see #133 for data)
  - The [code](https://github.com/Team-RADDISH/ParticleDA.jl/blob/tk/optimal-filter/src/OptimalFilter.jl) is tested and validated against [the original](https://github.com/Team-RADDISH/mini_apps/blob/master/optimal_particle_filter/Sample_Optimal_Height_Proposal.jl)
- [x] Integrate into ParticleDA.jl
  - [x] `sample_height_proposal` is called instead of adding the model noise by `add_random_field` —> everything else remains the same
    - This is difficult conceptually because the model noise is defined by the model in the current design. However `sample_height_proposal` should function independently of the model, ie. be a part of the filter.
    - It also needs access to `stations` and `observations` that are currently private fields of the model
    - [x] Double check with @AlexandrosBeskos that this interpretation is correct
    - [x] Re-design the model-filter interface to accommodate the optimal filter
  - [x] Add new `get_log_weights` function
    - This looks almost the same as our existing `get_log_weights` function but the covariance matrix is not hermitian. Can they be merged?

tkoskela:
## Plan to integrate into `ParticleDA.jl`
- [ ] Split `update_particles!()` into two parts
  1. `tsunami_update!()` + `add_random_field!()`
  2. `get_obs!()`
- [ ] Add a `filter_type` parameter to `run_particle_filter()` and Dispatch a different method based on the selected filter type 
- [ ] With optimal filter, update particles with `sample_height_proposal!()` between the two parts of `update_particles!()`
  - Expose the necessary fields of the model to the filter (possibly only when optimal filter is used)
    - [ ] read/write particles
    - [x] read observations
    - [ ] read stations

tkoskela:
**What do we do with the velocity fields once we've applied the sampling algorithm to the height?**

As far as I've understood, this is what we do in a loop:

1. Update height and velocities of all particles using the forward model PDE solver
2. Apply the sampling algorithm on the height field (of all particles) only --> new height field with optimal proposal

[...]

PR #145, opened 2021-01-27
Update model interface for optimal filter

tkoskela:
As planned in #136 

Split `update_particles!` to `update_particle_dynamics!` and `update_particle_noise!`

DONE: Now both new functions return the observations which may not be ideal. Would it be better to separate `get_particle_observations` into a third function?


giordano:
I pushed this last week https://github.com/Team-RADDISH/ParticleDA.jl/commit/28feb5234a87e8c3514e6a4ea741758bbc77cbe2 but forgot to open the PR :facepalm:  Yours is better, it has also tests! :smiley:

tkoskela:
That's exactly the thing in my TODO! Can we merge it into this PR?

tkoskela:
Yes we can! 💪

docs/src/index.md
```diff
@@ -68,6 +68,7 @@ be used by [`run_particle_filter`](@ref):
 ParticleDA.get_particles
 ParticleDA.get_truth
 ParticleDA.update_truth!
+ParticleDA.get_observations!
 ParticleDA.update_particles!
```
giordano, line 72: I think you need to do
```suggestion
ParticleDA.update_particle_dynamics!
ParticleDA.update_particle_noise!
ParticleDA.get_particle_observations!
```

tkoskela:
I'll mark this as ready for review. The other thing to do in #136 is expose more fields from the model_data struct but we can do that in another PR.

giordano (approved):

PR #146, opened 2021-02-11
Add two new functions to implement in the model

giordano:
The functions are used to get the size of the data and the number of state
variables, to initialise some arrays in the particle filter


tkoskela (approved):
`get_size` sounds a bit ambiguous. Could we call it something more specific? For example `get_grid_size`

giordano:
Yeah, I'm not attached at all to the names, happy to change them!

PR #148, opened 2021-02-19
Fix bug in update_reference_data.jl

tkoskela:
Change type to type instance in arguments of `run_particle_filter`


giordano (approved):
For a few seconds I wondered how tests could have worked so far without this :sweat_smile:

PR #153, opened 2021-02-22
Add GitHub Action workflow to run benchmarks on CI for pull requests

giordano:


giordano:
It works:
https://github.com/Team-RADDISH/ParticleDA.jl/blob/ceec818fcd6ea3b07c6f9e5abfb63d68fd0ab582/2021/02/22/164051/result.md :partying_face:  As I suspected, `benchmark/benchmarks.jl` should be already committed to `master` for this to work

giordano:
Of the benchmark results [published yesterday](https://github.com/Team-RADDISH/ParticleDA.jl/tree/1048bf6cc16a4426845a4101b5feb00fac1fd93d/2021/02/22), only the first and the last seems meaningful (no performance change between the baseline on master and this PR), while the other two report a 50% speedup (with no changes to the code...).

I'm not sure what's going on, I've run the `BenchmarkCI` code locally _**multiple**_ times and always found consistent results between the target and the baseline.  I suspect that running benchmark on GitHub-hosted machines may not be the best idea possible (perhaps machine are randomly hit by other load?).  In the past I tried to run benchmarks on Travis and results were unreliable at best, but "totally meaningless" is a better depiction of my experience, performance was extremely poor overall.  Azure/GitHub machines are better than Travis, but may still be not always reliable.

giordano:
Not sure this explains the issue (especially, not the difference between target and baseline), but the two "good runs" were using Intel(R) Xeon(R) Platinum 8171M CPU @ 2.60GHz, while the other two were on Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz.

tkoskela (approved):

PR #154, opened 2021-02-25
Add more benchmarks

giordano:


benchmark/benchmarks.jl
```diff
@@ -23,7 +33,22 @@ const params = Dict(
...
+SUITE["base"]["update_truth!"] = @benchmarkable ParticleDA.update_truth!($(model_data), $(nprt_per_rank))
+SUITE["base"]["update_particle_noise!"] = @benchmarkable ParticleDA.update_particle_noise!($(model_data), $(nprt_per_rank))
+SUITE["base"]["get_log_weights!"] = @benchmarkable ParticleDA.get_log_weights!(weights, truth_observations, model_observations, $(filter_params.weight_std)) setup=(weights = Vector{Float64}(undef, nprt_per_rank); truth_observations=ParticleDA.update_truth!(model_data, nprt_per_rank); model_observations = ParticleDA.get_particle_observations!(model_data, nprt_per_rank))
+SUITE["base"]["normalized_exp!"] = @benchmarkable ParticleDA.normalized_exp!(weights) setup=(weights = rand(filter_params.nprt))
+SUITE["base"]["resample!"] = @benchmarkable ParticleDA.resample!(resampling_indices, weights) setup=(resampling_indices = Vector{Int}(undef, filter_params.nprt); weights = rand(filter_params.nprt))
+# SUITE["base"]["copy_states!"] = @benchmarkable ParticleDA.copy_states!($(ParticleDA.get_particles(model_data)), $(bootstrap_filter_data.copy_buffer), $(bootstrap_filter_data.resampling_indices), $(my_rank), $(nprt_per_rank))
```
giordano, line 48: This is failing inside `MPI.Irecv!` but I don't understand why.  Not sure it's worth spending more time on this, but I left it commented in case someone wants to get back to this in the future.

benchmark/benchmarks.jl
```diff
@@ -1,10 +1,20 @@
...
+if !MPI.Initialized()
+    MPI.Init()
+end
+
 const SUITE = BenchmarkGroup()
+const my_rank = MPI.Comm_rank(MPI.COMM_WORLD)
+const my_size = MPI.Comm_size(MPI.COMM_WORLD)
+
+SUITE["base"] = BenchmarkGroup()
+SUITE["BootstrapFilter"] = BenchmarkGroup()
```
giordano, line 17: Now we have these two benchmark groups: `"base"` is common to all filters (I think, let me know if I missed something!), `"BootstrapFilter"` runs benchmarks for `init_filter` and `run_particle_filter` which in #133 will have different implementations for the optimal filters

tkoskela: What about `update_particle_dynamics!`?

giordano: Good catch!

tkoskela (approved):

PR #155, opened 2021-02-26
Pass filter type to run_particle_filter and use it to dispatch 

tkoskela:
Seems to work, but maybe better to have a look with a second 👀


src/ParticleDA.jl
```diff
@@ -362,29 +362,46 @@ function init_filter(filter_params::FilterParameters, model_data, nprt_per_rank:
...
+        # Add noise from the standard gaussian random field to all state variables in model_data
+        update_particle_noise!(model_data, nprt_per_rank)
+
+        # Overwrite the height state variable with the samples of the optimal proposal
+        particles[:,:,1,:] .= filter_data.online_matrices.samples
+        set_particles!(model_data, particles)
+    
+end
+
+function run_particle_filter(init, filter_params::FilterParameters, model_params_dict::Dict, filter_type)
```
giordano, line 404: I'd annotate the last argument with the type
```suggestion
function run_particle_filter(init, filter_params::FilterParameters, model_params_dict::Dict, filter_type::ParticleFilter)
```

giordano (approved):
Looks good beside the type annotation

PR #157, opened 2021-03-01
Split get_grid_size into three functions 

tkoskela:
Each function returns a NTuple so that they can be used for higher dimensional grids. This will address comments in #133

### DIFF
--- a/src/OptimalFilter.jl
+++ b/src/OptimalFilter.jl
@@ -36,6 +36,15 @@ struct OnlineMatrices{T<:AbstractVector, S<:AbstractArray, U<:AbstractArray}
 
 end
 
+struct Grid{T}
+    nx::Int
+    ny::Int
+    dx::T
+    dy::T
+    x_length::T
+    y_length::T
+end
+
 # Covariance function r(x,y), equation 1 in Dietrich and Newsam 96
 function covariance(x::T, y::T, noise_params::NamedTuple) where T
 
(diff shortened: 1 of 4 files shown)


src/ParticleDA.jl
```diff
@@ -14,13 +14,23 @@ using .Default_params
 # Functions to extend in the model
 
 """
-    ParticleDA.get_grid_size(model_data) ->
-        NamedTuple({:nx, :ny, :dx, :dy, :x_length, :y_length},Tuple{Int, Int, Float, Float, Float, Float})
+    ParticleDA.get_grid_dims(model_data) -> NTuple{N, Int} where N
```
giordano, line 17:
```suggestion
    ParticleDA.get_grid_size(model_data) -> NTuple{N, Int} where N
```

src/OptimalFilter.jl
```diff
@@ -44,7 +44,7 @@ function covariance(x::T, y::T, noise_params::NamedTuple) where T
 end
 
 # Extended covariance function /bar r(x,y), equation 8 of Dietrich and Newsam 96
-function extended_covariance(x::T, y::T, grid::NamedTuple, noise_params::NamedTuple) where T
+function extended_covariance(x::T, y::T, grid, noise_params::NamedTuple) where T
```
giordano, line 47: It's probably better to annotate
```suggestion
function extended_covariance(x::T, y::T, grid::Grid, noise_params::NamedTuple) where T
```
here and in all other definitions below.  This is more to make sure we call the function with the right argument than else (no performance gain)

giordano (approved):

PR #158, opened 2021-03-01
More generic initialisation of boostrap filter

giordano:
I tried to improve genericity of the init method.  The sizes of the arrays
aren't hard-coded but depends on the dimensionality of model data.  I ran the
benchmarks locally a few times and didn't see any regression, which in any case
should be limited to the init method, which acts as a function barrier.  Nothing
should change at runtime.


tkoskela (approved):
Looks good!

PR #159, opened 2021-03-04
Precompute FFTW plan

giordano:
Very work-in-progress (still needs to adapt tests)

### DIFF
--- a/src/OptimalFilter.jl
+++ b/src/OptimalFilter.jl
@@ -140,40 +140,54 @@ end
 # Normalized two-dimension discrete Fourier transofrm normalized by sqrt(n1_bar).
 # Operates on the 2d data stored as a matrix.
 # From Dietrich & Newsam 96 in text following equation 12
-function normalized_2d_fft!(transformed_array::AbstractMatrix{<:Complex}, array::AbstractMatrix{S}, fft_plan::FFTW.FFTWPlan, grid_ext) where {S}
+function normalized_2d_fft!(transformed_array::AbstractMatrix{<:Complex}, array::AbstractMatrix{S}, fft_plan::FFTW.FFTWPlan, fft_plan!::FFTW.FFTWPlan, grid_ext) where {S}
 
     normalization_factor = 1.0 / sqrt(grid_ext.nx * grid_ext.ny)
-    transformed_array .= (fft_plan * complex(array)) .* normalization_factor
+    if pointer(transformed_array) == pointer(array)
+        mul!(transformed_array, fft_plan!, array)
+    else
+        mul!(transformed_array, fft_plan, array)
+    end
+    transformed_array .*= normalization_factor
 
 end
 
 # Normalized two-dimension discrete Fourier transofrm normalized by sqrt(n1_bar).
 # Operates on the 2d data stored as a vector.
 # From Dietrich & Newsam 96 in text following equation 12
-function normalized_2d_fft!(transformed_vector::AbstractVector{<:Complex}, vector::AbstractVector{S}, fft_plan::FFTW.FFTWPlan, grid_ext) where {S}
+function normalized_2d_fft!(transformed_vector::AbstractVector{<:Complex}, vector::AbstractVector{S}, fft_plan::FFTW.FFTWPlan, fft_plan!::FFTW.FFTWPlan, grid_ext) where {S}
 
     normalization_factor = 1.0 / sqrt(grid_ext.nx * grid_ext.ny)
-    transformed_vector .= (fft_plan * reshape(vector, grid_ext.nx, grid_ext.ny))[:] .* normalization_factor
+    tmp_array = complex(reshape(vector, grid_ext.nx, grid_ext.ny))
+    mul!(tmp_array, fft_plan!, tmp_array)
+    transformed_vector .= @view(tmp_array[:]) .* normalization_factor
 
 end
 
 # Normalized inverse two-dimension discrete Fourier transofrm normalized by sqrt(n1_bar).
 # Operates on the 2d data stored as a matrix.
 # From Dietrich & Newsam 96 in text following equation 12
-function normalized_inverse_2d_fft!(transformed_array::AbstractMatrix{<:Complex}, array::AbstractMatrix{S}, fft_plan::FFTW.FFTWPlan, grid_ext) where {S}
+function normalized_inverse_2d_fft!(transformed_array::AbstractMatrix{<:Complex}, array::AbstractMatrix{S}, fft_plan::FFTW.FFTWPlan, fft_plan!::FFTW.FFTWPlan, grid_ext) where {S}
 
     normalization_factor = sqrt(grid_ext.nx * grid_ext.ny)
-    transformed_array .= (inv(fft_plan) * complex(array)) .* normalization_factor
+    if pointer(transformed_array) == pointer(array)
+        mul!(transformed_array, inv(fft_plan!), array)
+    else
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark.yml
```diff
@@ -28,7 +28,8 @@ jobs:
                   env = Dict(
                       "JULIA_NUM_THREADS" => "2",
                   ),
-              )
+              ),
+              "origin/tk/optimal-filter",
```
giordano, line 32: This needs to be reverted before merging, I used it to benchmark against the optimal filter branch

giordano:
Ready for review but not for merging, I'll need to revert the change to the benchmark before that

src/OptimalFilter.jl
```diff
@@ -139,59 +139,64 @@ end
...
-    normalization_factor = 1.0 / sqrt(grid_ext.nx * grid_ext.ny)
-    transformed_vector .= fft(reshape(vector, grid_ext.nx, grid_ext.ny))[:] .* normalization_factor
-
-end
-
-# Normalized inverse two-dimension discrete Fourier transofrm normalized by sqrt(n1_bar).
-# Operates on the 2d data stored as a matrix.
-# From Dietrich & Newsam 96 in text following equation 12
-function normalized_inverse_2d_fft!(transformed_array::AbstractMatrix{T}, array::AbstractMatrix{S}, grid_ext) where {T,S}
+function normalized_2d_fft!(transformed_vector::AbstractVector{<:Complex}, vector::AbstractVector{S}, fft_plan::FFTW.FFTWPlan, fft_plan!::FFTW.FFTWPlan, grid_ext, f=identity) where {S}
```
tkoskela, line 162: `fft_plan` is not used?

src/OptimalFilter.jl
```diff
@@ -139,59 +139,64 @@ end
...
-function normalized_2d_fft!(transformed_array::AbstractMatrix{T}, array::AbstractMatrix{S}, grid_ext) where {T,S}
+function normalized_2d_fft!(transformed_array::AbstractMatrix{<:Complex}, array::AbstractMatrix{S}, fft_plan::FFTW.FFTWPlan, fft_plan!::FFTW.FFTWPlan, grid_ext, f=identity) where {S}
 
-    normalization_factor = 1.0 / sqrt(grid_ext.nx * grid_ext.ny)
-    transformed_array .= fft(array) .* normalization_factor
+    normalization_factor = f(sqrt(grid_ext.nx * grid_ext.ny))
+    if pointer(transformed_array) == pointer(array)
+        mul!(transformed_array, f(fft_plan!), array)
+    else
+        mul!(transformed_array, f(fft_plan), array)
```
tkoskela, line 151: What's the difference between `fft_plan` and `fft_plan!`?

src/OptimalFilter.jl
```diff
@@ -139,59 +139,64 @@ end
...
+# The argument `f` lets you switch between forward transform (`f=identity`) and
+# inverse transform (`f=inv`).
 # From Dietrich & Newsam 96 in text following equation 12
-function normalized_2d_fft!(transformed_array::AbstractMatrix{T}, array::AbstractMatrix{S}, grid_ext) where {T,S}
+function normalized_2d_fft!(transformed_array::AbstractMatrix{<:Complex}, array::AbstractMatrix{S}, fft_plan::FFTW.FFTWPlan, fft_plan!::FFTW.FFTWPlan, grid_ext, f=identity) where {S}
 
-    normalization_factor = 1.0 / sqrt(grid_ext.nx * grid_ext.ny)
-    transformed_array .= fft(array) .* normalization_factor
+    normalization_factor = f(sqrt(grid_ext.nx * grid_ext.ny))
+    if pointer(transformed_array) == pointer(array)
```
tkoskela, line 148: What are the pointers for?

src/OptimalFilter.jl
```diff
@@ -139,59 +139,64 @@ end
...
-# Operates on the 2d data stored as a vector.
 # From Dietrich & Newsam 96 in text following equation 12
-function normalized_inverse_2d_fft!(transformed_vector::AbstractVector{T}, vector::AbstractVector{S}, grid_ext) where {T,S}
+function normalized_inverse_2d_fft!(transformed_array::AbstractArray{<:Complex}, array::AbstractArray{S}, fft_plan::FFTW.FFTWPlan, fft_plan!::FFTW.FFTWPlan, grid_ext) where {S}
 
-    normalization_factor = sqrt(grid_ext.nx * grid_ext.ny)
-    transformed_vector .= ifft(reshape(vector, grid_ext.nx, grid_ext.ny))[:] .* normalization_factor
+    return normalized_2d_fft!(transformed_array, array, fft_plan, fft_plan!, grid_ext, inv)
 
 end
```
tkoskela, line 177: Do we need this function if we can just call `normalized_2d_fft!` with `inv`?

src/ParticleDA.jl
```diff
@@ -366,12 +366,16 @@ function init_filter(filter_params::FilterParameters, model_data, nprt_per_rank:
 
     model_noise_params = get_model_noise_params(model_data)
     obs_noise_std = get_obs_noise_std(model_data)
+    # Precompute two FFT plans, one in-place and the other out-of-place
+    C = complex(T)
+    tmp_array = Matrix{C}(undef, grid_ext.nx, grid_ext.ny)
+    fft_plan, fft_plan! = FFTW.plan_fft(tmp_array), FFTW.plan_fft!(tmp_array)
 
-    offline_matrices = init_offline_matrices(grid, grid_ext, stations, model_noise_params, obs_noise_std, T)
+    offline_matrices = init_offline_matrices(grid, grid_ext, stations, model_noise_params, obs_noise_std, fft_plan, fft_plan!, T)
```
tkoskela, line 374: It looks like `fft_plan, fft_plan!` are always passed together so could just store them as a tuple?

src/OptimalFilter.jl
giordano: The out-of-place `fft_plan` can be used only in
```julia
mul!(B, fft_plan, A)
```
with `A` and `B` different arrays, or more precisely `pointer(A) != pointer(B)`: https://github.com/JuliaMath/FFTW.jl/blob/2c20abe7098a2380fa37fe021b3d21fc593059ec/src/fft.jl#L438.  Instead, the in-place `fft_plan!` can only be used in
```julia
mul!(A, fft_plan!, A)
```
We do have cases where we overwrite the input array.

src/OptimalFilter.jl
giordano: See the answer to https://github.com/Team-RADDISH/ParticleDA.jl/pull/159#discussion_r589433961

src/OptimalFilter.jl
giordano: That's correct, but all these methods need to have the same signature in terms of number of arguments.

src/OptimalFilter.jl
giordano: That's a good point, I'm going to remove this function

src/ParticleDA.jl
giordano: This is actually what I did at the beginning, but I didn't like it too much because it was opaque what `fft_plan[1]` and `fft_plan[2]` were (which is the out-of-place, which is the in-place?).  I guess you'd suggest a named tuple? :smile:

tkoskela: 😁  Not a big deal either way

PR #160, opened 2021-03-08
Add integration test with optimal filter

tkoskela:
Added an integration test with optimal filter. Should be ok using StableRNG but I get failures intermittently for no obvious reason. Adding `Random.seed!(rng, 123)` seems to help, but I don't exactly understand why.


giordano:
Uhm, we may need to regenerate the reference data

src/OptimalFilter.jl
```diff
@@ -363,10 +363,13 @@ function sample_height_proposal!(height::AbstractArray{T,3},
...
 
+    e1 = Vector{ComplexF64}(undef, grid_ext.nx * grid_ext.ny)
+    e2 = Vector{ComplexF64}(undef, stations.nst)
+
     for iprt in 1:2:filter_params.nprt
         # TODO we could pre-create all our random numbers in one go before the loop, would that be faster?
-        e1 = complex.(randn(rng, grid_ext.nx*grid_ext.ny), randn(rng, grid_ext.nx*grid_ext.ny))
-        e2 = complex.(randn(rng, stations.nst), randn(rng, stations.nst))
+        @. e1 = complex(randn(rng), randn(rng))
+        @. e2 = complex(randn(rng), randn(rng))
```
giordano, line 372: With this simple change (allocate `e1` and `e2` once before the loop, overwrite them inside the loop) we cut memory allocations by 30% and improve runtime by 18% in `run_particle_filter`: https://github.com/Team-RADDISH/ParticleDA.jl/blob/bfb6ef5ef19528979c07a71a2c01d861b3717120/2021/03/11/204257/result.md#results :slightly_smiling_face:

giordano: However I'm going to split performance improvements in a different PR, so it's easier to follow the changes here about the tests

PR #167, opened 2021-03-30
Add metadata to StateVectors with FieldMetadata.jl.

### LINKED ISSUE #165: Organise model metadata
The metadata for most of the model output is just ad-hoc hard-coded into the I/O routines, e.g. 

https://github.com/Team-RADDISH/ParticleDA.jl/blob/acf19df0f786c46bce2a4694b2ebf707c6ec105b/test/model/model.jl#L599

This should be added to the data structures and read from there. We could look into [FieldMetadata.jl](https://github.com/rafaqz/FieldMetadata.jl), looks like it does exactly that.

tkoskela:
Remove hard-coded metadata from write_snapshot().

Not perfect but I'd say it closes #165


giordano (approved):
Always nice to see diffs with 45 additions and 57 deletions :smiley:

PR #176, opened 2021-04-23
Support different noise parameters for each state variable

### LINKED ISSUE #175: Different Background Noise Parameters for Each of Three State Variables
`lambda`, `sigma` and `nu` https://github.com/Team-RADDISH/ParticleDA.jl/blob/afd102ba5529f629a0a5f227a8142601050dfab8/test/model/model.jl#L85-L87

Should have one value per state variable. So in this example three values, one for height, vx, vy. They could be tuples, or whatever is most appropriate.

When we sample the random field https://github.com/Team-RADDISH/ParticleDA.jl/blob/afd102ba5529f629a0a5f227a8142601050dfab8/test/model/model.jl#L291

We should not use the same `generator` for each one, but rather we should initialise a generator for each `(lambda[ivar], sigma[ivar], nu[ivar])` and then use `generator[ivar]` here

tkoskela:
fixes #175

### DIFF
--- a/src/ParticleDA.jl
+++ b/src/ParticleDA.jl
@@ -51,7 +51,7 @@ function get_obs_noise_std end
 """
     ParticleDa.get_model_noise(model_data) -> NamedTuple({:sigma, :lambda, :nu},Tuple{Float, Float, Float})
 
-Return standard deviation of observation noise. Required for optimal filter only.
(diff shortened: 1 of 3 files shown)


giordano (approved):
I slightly changed how the random fields are generated for each component in order to return a vector with a concrete type, the rest looks good

PR #183, opened 2021-05-11
Add Julia plotting notebook

### LINKED ISSUE #182: Plotting tool in Julia
To simplify user experience, we should have a plotting tool in Julia. Functionality can follow 

https://github.com/Team-RADDISH/ParticleDA.jl/blob/master/extra/Plot_tdac_output.ipynb

giordano:
Fix #182


tkoskela:
One comment actually, should we add a `particle_da.h5` file to the repository so this would plot something immediately after cloning the repo?

tkoskela:
And another one, the name of both this and the python notebook should not mention tdac 😅

PR #186, opened 2021-05-20
Remove redundant call to `get_particles`

giordano:


tkoskela (approved):
For my peace of mind, the `=` sign here

https://github.com/Team-RADDISH/ParticleDA.jl/blob/a54959bc8baf6718dde5eaeceded18ad2abff3fc/test/model/model.jl#L432

creates a reference to `d.states.particles`?

giordano:
> For my peace of mind, the `=` sign here
>
> https://github.com/Team-RADDISH/ParticleDA.jl/blob/a54959bc8baf6718dde5eaeceded18ad2abff3fc/test/model/model.jl#L432
>
> creates a reference to `d.states.particles`?

Yes, this function here basically returns a reference to the `d.states.particles` array, so when you do

```julia
particles = ParticleDA.get_particles(model_data)
```

the `particles` variable points to the memory address of `d.states.particles`, and if you modify the `particles` variable you're actually modifying `d.states.particles`.

> Now that I think about it, that makes is possible for the model implementation to break things if it doesn't return a reference.

The only way I can see this to be possible is if a `ParticleDA.get_particles` method returns a copy of the actual particles, rather than the array itself, which I think is a weird thing to do.

> Should we have some kind of a check for that?

I'd be happy to add a test, but not sure what to test :thinking:  Maybe we should better document this?

tkoskela:
Let's add it in the docstring of `get_particles` (and others if applicable) that it should return a reference. That's probably good enough for now.

tkoskela:
A simple test off the top of my head would be

```
particles=get_particles(d)
particles .= 0.0
particles=get_particles(d)
assert particles \approx 0.0
```

giordano:
Good idea!  I improved docstring of `get_particles` to specify it needs to return the vector of particles because that will need to be modified in-place, and for the test I went with something similar to but more fundamental than what you suggested: the pointer of `get_particles(model_data)` is always the same

PR #188, opened 2021-05-21
Update weak scaling scripts on CSD3

tkoskela:
Leaving as draft because the julia 1.5.2 module I'm loading gives errors on illegal instructions. These run fine if I use a locally installed julia binary, but that may not be what we should do in master. I guess I need to ask the CSD3 people to update their julia modules.


tkoskela:
Closing this as outdated

Issue #199, opened 2021-09-08
Add SPEEDY climate model


tkoskela:
Within the extension of the project, we are adding a new model, [speedy](https://github.com/samhatfield/letkf-speedy/tree/master/model)

## Workflow
- Call speedy bash script that
  - sets up inputs
  - executes Fortran executable
- Read outputs from Fortran binary files and populate `model_data`

## Tasks
- [x] write new `update_particle_dynamics!` function
- [ ] write IO function(s) for speedy output files

tkoskela:
Issues with 2D vs 3D
- Gaussian random field generator: Looks like it will work but needs testing
- get_mean_and_var: Easy to generalise because we don't use the dimension. We can assert that `size(statistics) == size(particles) - 1`
- copy_states: Needs to be generalised to use the "last" dimension. [This should do it](https://github.com/ChrisRackauckas/EllipsisNotation.jl)
- State vectors: Are defined in the model

DanGiles:
I am having difficulty with writing to a Fortran file that SPEEDY will accept. Please see my current [write_fortran](https://github.com/DanGiles/ParticleDA.jl/blob/dg/speedy/speedy/model/model.jl) (line 520) function which relies on FortranFiles.jl. The files it creates can be read in correctly using our previous reading function. Unfortunately, SPEEDY will not accept them and they cause the SPEEDY model to crash - with no useful outputs appearing. 

The write_fortran function should replicate the behaviour of the following SPEEDY [write_grd4](https://github.com/samhatfield/letkf-speedy/blob/master/common/common_speedy.f90) function (line 197). 

Perhaps this behaviour is due to the `unformatted` argument (padding the records before and after) in write_grd4 currently not being replicated in the Julia function. Your help would be much appreciated!

giordano:
Could you please show exactly what you've tried?  For example, I tried to generate a file with your `write_fortran` function like this:
```julia
using .Model: write_fortran

nlon = 2
nlat = 
nlev = 2
filename = tempname()
nij0 = nlon*nlat
iolen = 4
T = Float32
dataset = rand(T, nlon, nlat, nlev, 6)
write_fortran(filename, nlon, nlat, nlev, dataset)
```
and then I can read it back, always using the Julia package `FortranFiles`, with
```julia
f = FortranFile(filename, "r", convert="big-endian", access="direct", recl=(nij0*iolen))
read(f, T; rec=1)
```
[...]

PR #201, opened 2021-10-18
Generalise some dimensions in bootstrap filter

giordano:
Ref: https://github.com/Team-RADDISH/ParticleDA.jl/issues/199#issuecomment-943327958.  CC @DanGiles

### DIFF
--- a/test/copy_states.jl
+++ b/test/copy_states.jl
@@ -7,34 +7,38 @@ my_size = MPI.Comm_size(MPI.COMM_WORLD)
 nprt_per_rank = 3
 n = nprt_per_rank * my_size
 
-local_particles = float(collect(my_rank * nprt_per_rank + 1 : (my_rank + 1) * nprt_per_rank))
+@testset for size in ((1,1,1,nprt_per_rank), (1,1,1,1,nprt_per_rank))
 
-for i = 1:my_size
-    if i == my_rank + 1
-        println("rank ",my_rank,": particle states: ", local_particles)
+    local_particles = float(collect(my_rank * nprt_per_rank + 1 : (my_rank + 1) * nprt_per_rank))
+
+    for i = 1:my_size
+        if i == my_rank + 1
+            println("rank ",my_rank,": particle states: ", local_particles)
+        end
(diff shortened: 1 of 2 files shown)


DanGiles:
Unfortunately, I'm having difficulties with the latest merged version. MPI is throwing an error when I run:
`run_particle_filter(Model.init, input_file, BootstrapFilter())` 

`Fatal error in MPI_Init: Other MPI error, error stack:
MPI_Init(101): Cannot call MPI_INIT or MPI_INIT_THREAD more than once
[unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=805363215
:
system msg for write_line failure : Bad file descriptor
`

MPI does not appear in the list of precompiled when I run `Pkg.instantiate()`, therefore I think the issue is related to the Project.toml or Manifest.toml file?

giordano:
I don't think that's the error, otherwise you'd have a get a much different error (e.g. `UndefVarError: MPI not defined`) when trying to call that function.  That error means that you can't call `MPI.Init()` multiple times in a session.  Note: this is a fundamental limitation of the MPI standard, it doesn't have anything to do with Julia.

However, I presume you are testing in #198?  I noticed you [downgraded](https://github.com/Team-RADDISH/ParticleDA.jl/pull/198/files#diff-fbcf94675bcaa4000636848fd45149c8ac6588d10bc347563ccf6b8faca1d368R25) the version of `MPI.jl`, I'm not sure why you did that, but note that version v0.19 `MPI.jl` allows calling `MPI.Init()` multiple times in a session without problems because it internally checks whether MPI is already initialised, instead of us having to guard all `MPI.Init` calls like this
https://github.com/Team-RADDISH/ParticleDA.jl/blob/0ad9943a02fa52e47c781011a9fd66b2dc00de69/src/ParticleDA.jl#L416-L418
It's likely you are now somewhere calling `MPI.Init()` multiple times with an older version of `MPI.jl` and without the appropriate guard.

DanGiles:
Thanks Mosé. I forked the code at an earlier point and then forgot to merge with the upstream. 
I will correct this now and use v0.19 MPI.jl

PR #210, opened 2022-01-18
Use correct number of particles in Optimal Filter and add MPI test

tkoskela:
Optimal Filter does not do any global operations over particles, so we can simply substitute `nprt_per_rank` for `filter_params.nprt`. This has an added benefit of not having to pass `filter_params` at all to most functions, since `nprt` was the only field being used.

Optimal Filter is now included in the tests in `mpi.jl`, so we should at least catch MPI errors a bit sooner.

Also removed verbose output from the MPI test, didn't see any reason to have it. Unless someone disagrees? The MPI test also enables timers, do we want to keep that?

Closes #209

PR #211, opened 2022-01-18
Propagate number of threads when running MPI tests

giordano:
While looking at new failures in #210 I realised we always use a single in MPI
tests because we don't forward the number of threads to the `julia` subprocess
we spawn.  I think it'd be good to run the tests with multiple MPI ranks +
multiple threads.


tkoskela:
Now that I think about it, the `-t` should probably have gone into `flags`

giordano:
Ugh, excellent point: #212

PR #216, opened 2022-01-27
Use `GaussianRandomFields` to calculate optimal proposal covariance

### LINKED ISSUE #213: Covariance function assumed in optimal proposal not guaranteed to match that used for process noise
Currently the covariance function (in the Gaussian process / random field sense) used for computing the mean and covariance of the (Gaussian) optimal proposal is hardcoded in the following

https://github.com/Team-RADDISH/ParticleDA.jl/blob/c278bbc46db25cb517f58471198a7bb189b1fc3f/src/OptimalFilter.jl#L48-L53

Separately each model is expected to define a `update_particle_noise!` method which perturbs the state particles (post propagating forwards in time under the deterministic model dynamics) with additive Gaussian state / process noise. In both the currently implemented TDAC model in `test/model` and in @DanGiles' ongoing implementation of the SPEEDY model in #203 the `update_particle_noise!` methods use the `GaussianRandomFields` package to sample a Gaussian random field to use as the additive Gaussian state noise. `GaussianRandomFields` itself provides methods for evaluating a covariance function at [one](https://github.com/PieterjanRobbe/GaussianRandomFields.jl/blob/d0c3e9e3d6e70f39c8d2d03cb25d0cc4579f8fe0/src/covariance_functions/covariance_functions.jl#L129) or [multiple](https://github.com/PieterjanRobbe/GaussianRandomFields.jl/blob/d0c3e9e3d6e70f39c8d2d03cb25d0cc4579f8fe0/src/covariance_functions/covariance_functions.jl#L132) pairs of points. 

The current `covariance` implementation in `src/OptimalFilter.jl` hardcodes both the family of covariance function and the specific parameterization used. If a model uses a covariance function from a different family or from the same family but with a different parameterization then the state noise covariances computed to construct the optimal proposal and in simulating the state noise will not match and lead to incorrect results. 

If the `covariance` implementation in `src/OptimalFilter.jl` was instead passed the `CovarianceStructure` struct used to define the state noise Gaussian random field, and used [the `apply` method](https://github.com/PieterjanRobbe/GaussianRandomFields.jl/blob/d0c3e9e3d6e70f39c8d2d03cb25d0cc4579f8fe0/src/covariance_functions/covariance_functions.jl#L129) from `GaussianRandomFields` internally, this would both ensure consistency between the definitions without having to worry about the particular parameterization used and also allow the implementation to straight away generalize to any of the other stationary covariance functions defined in `GaussianRandomFields`.

As a related point, it is not entirely clear if the covariance function definition for the TDAC model does match that in the optimal proposal implementation currently. In

https://github.com/Team-RADDISH/ParticleDA.jl/blob/c278bbc46db25cb517f58471198a7bb189b1fc3f/test/model/model.jl#L250-L272

the state noise Gaussian random field (for the fields for each of the three state variables) is set to have a Matern covariance function with lengthscale parameter `lambda = 1e4`, smoothness parameter `nu = 2.5` and marginal standard deviation `sigma = 1`.  The [documentation for the Matern covariance function in the `GaussianRandomFields` package](https://pieterjanrobbe.github.io/GaussianRandomFields.jl/dev/API/#GaussianRandomFields.Matern) suggests the corresponding covariance function definition is

```
r(x_1, x_2) = sigma * (2^(1 - nu)) / (gamma(nu)) * (rho(x_1, x_2) / lambda)^nu * besselk(nu, rho(x_1, x_2) / lambda)
```

with `rho(x_1, x_2) = sum(abs(x_1 .- x_y).^2)`. While the Matern covariance (under a different parametrization) does [simplify for certain values of the smoothness parameter `nu`](https://en.wikipedia.org/wiki/Mat%C3%A9rn_covariance_function#Simplification_for_%CE%BD_half_integer) it is not clear this matches the corresponding defnition in `OptimalFilter.jl`

```
r(x_1, x_2) = sigma^2 * exp(−(sum(abs(x_1 .- x_2)) / (2 * lambda))
```

### LINKED ISSUE #214: Fix TODO about covariance in comment
The answer to the question in 

https://github.com/Team-RADDISH/ParticleDA.jl/blob/c278bbc46db25cb517f58471198a7bb189b1fc3f/src/OptimalFilter.jl#L114-L122

is I believe explained in the first complete paragraph in the second page (page number 1645) of [Dietrich and Newsam 1995](https://doi.org/10.1029/94WR02977)

> Second, in practice, measurements z₂* of Z  on Ω are likely to be noisy. This is usually modelled by assuming that [z₂*]ₖ = Z(xₖ, yₖ) + εₖ, where the errors εₖ, are zero mean Gaussian random variables which may be correlated among themselves with correlation matrix Σ, but are independent of the vectors z₁ and z₂. This extension can be easily accommodated in the results presented above by simply replacing the matrix R₂₂ in (3) and subsequent equations by R̅₂₂ =  R₂₂ + Σ. 

In our case we are assuming Σ = σ²I (that is the observation errors / noise variables are independent zero-mean Gaussian random variables with standard deviation σ) so R̅₂₂ =  R₂₂ + σ²I. 

If that answers the question I can create a PR to update the comment to summarize the above.

matt-graham:
Fixes #213 and fixes #214.

Updates the `covariance` method in `src/OptimalFilter.jl` to use the `apply` method in the `GaussianRandomFields` package rather than hardcoding a particular covariance function. The expected interface of a model is also slightly updated to have `get_model_noise_params` return a structure defining the parameters of the covariance function. These changes both ensure the covariance function definition used in sampling the state noise for the model and assumed in the optimal proposal distribution match (which may not currently be the case as described in #213) and means the optimal proposal distribution directly generalises to any isotropic covariance function defined in the `GaussianRandomFields` package.

I also updated some of the comments in `src/OptimalFilter.jl` to clarify the match to the equations / notation in Dietrich and Newsam (1996), including specifically resolving the TODO described in #214.


src/OptimalFilter.jl
```diff
@@ -45,15 +46,13 @@ struct Grid{T}
...
 end
 
-# Covariance function r(x,y), equation 1 in Dietrich and Newsam 96
-function covariance(x::T, y::T, noise_params::NamedTuple) where T
-
-    return noise_params.sigma^2 * exp(-(abs(x) + abs(y))/(2 * noise_params.lambda))
-
+# Covariance function r(x,y), equation 1 in Dietrich & Newsam 1996
+function covariance(x::T, y::T, covariance_structure::IsotropicCovarianceStructure{T}) where T
+    return apply(covariance_structure, [x, y])
```
giordano, line 51: What's `apply`? :thinking:

matt-graham: Yes sorry should have made that more explicit.

matt-graham:
Turns out I had not being using my local development version of `ParticleDA.jl` when running the test locally 🤦 which unsurprisingly meant all the tests were passing. The failing unit tests I think just need me to update these lines

https://github.com/Team-RADDISH/ParticleDA.jl/blob/c278bbc46db25cb517f58471198a7bb189b1fc3f/test/runtests.jl#L296
https://github.com/Team-RADDISH/ParticleDA.jl/blob/c278bbc46db25cb517f58471198a7bb189b1fc3f/test/runtests.jl#L366

to reflect that `noise_params` is now expected to be a `GaussianRandomFields` covariance structure.

The optimal filter integration test also fails due to mismatch from the reference values. If there was a mismatch between the covariance being assumed in the state noise generation and optimal proposal distribution implementation that would possibly explain this. I'll check whether the values we are getting from the `covariance` method have changed, and if so regenerate the reference data.

matt-graham:
I've updated the tests to fix the problems described above, plus adjusting some manual covariance function definitions in `optimal_filter_validation.jl` and `runtests.jl` to be consistent with the Matern covariance function being used (specifically using the [simplified form of the Matern covariance for ν=0.5](https://en.wikipedia.org/wiki/Mat%C3%A9rn_covariance_function#Simplification_for_%CE%BD_half_integer) as the test case).

I also noticed what I think is a bug in `GaussianRandomFields` which means the σ parameter is being ignored when evaluating covariance matrices using `apply`. I have raised an issue about this: PieterjanRobbe/GaussianRandomFields.jl/issues/45. For now I have manually multiplied the output of `apply` by `σ^2` to give the correctly scaled covariances. If this is confirmed as a bug and fixed upstream in `GaussianRandomFields` we will need to remove this manual scaling.

test/optimal_filter_validation.jl
```diff
@@ -98,7 +98,7 @@ Sobs = 0.01
 #### Example Specification of Covariance Function
 function R(x::Float64, y::Float64, th)
 
-  Cov = (th.s^2)*exp(-(abs(x)+abs(y))/(2.0*th.l))
+  Cov = (th.s^2)*exp(-norm([x, y])/th.l)
```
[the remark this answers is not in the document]
matt-graham: > To be clear, is this one of the the adjustments you mentioned? (abs(x) + abs(y)) / 2 isn't quite the same as the sum of the squares of the two numbers 😅

Yeah - `GaussianRandomFields` [uses `norm` on the difference vector](https://github.com/PieterjanRobbe/GaussianRandomFields.jl/blob/239e28540367e0b020725d0e87b5fa7a8401dfec/src/covariance_functions/covariance_functions.jl#L129) (here `[x, y]`) as the `x` argument to the `apply` method for each subtype of `IsotropicCovarianceStructure` (for example [here](https://github.com/PieterjanRobbe/GaussianRandomFields.jl/blob/master/src/covariance_functions/matern.jl#L46-L52) for the Matern covariance), so I followed that here to give consistent values. The previous implementation I guess would have corresponded to a 1-norm so setting the optional `p` argument to `Matern` to 1 (which is passed to `norm`), with the 1/2 factor I think just a result of a different convention for the lengthscale parametrization.

Thanks for the tip about `hypot` - I hadn't known this existed, but a good one to have in the arsenal!

test/optimal_filter_validation.yaml
```diff
@@ -11,4 +11,4 @@ model:
     obs_noise_std: 0.01
     sigma: 1.0
     lambda: 1.0
-    nu: 2.5
+    nu: 0.5
```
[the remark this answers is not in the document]
matt-graham: This is what I was referring to with

> to be consistent with the Matern covariance function being used (specifically using the simplified form of the Matern covariance for ν=0.5 as the test case)

As the Matern covariance function is particularly simple for `nu=0.5` I changed to using this value so that in the `optimal_filter_validation.jl` definition of `R`
https://github.com/Team-RADDISH/ParticleDA.jl/blob/1f3e8e3ba331d4af00e2b802eb009489e5ff79ef/test/optimal_filter_validation.jl#L99-L105
we can just use a simple exponential expression rather than having to use `SpecialFunctions` to load the `gamma` and `besselk` functions. It might still be better to do this though I guess from a robustness standpoint so that the test is valid for any value of `nu`. Or at least I should perhaps put a comment in the YAML file to make clear this value shouldn't be changed?

src/ParticleDA.jl
```diff
@@ -50,9 +50,10 @@ Return standard deviation of observation noise. Required for optimal filter only
 function get_obs_noise_std end
 
 """
-    ParticleDa.get_model_noise(model_data) -> NamedTuple({:sigma, :lambda, :nu},Tuple{Float, Float, Float})
+    ParticleDa.get_model_noise(model_data) -> CovarianceStructure
```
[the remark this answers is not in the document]
matt-graham: Ah yes I forgot to update this, thanks for catching. I was originally trying to generalise `covariance` and `extended_covariance` in `OptimalFilter.jl` to work for both isotropic and non-isotropic covariance functions (as we would need this for the spherical distance based covariance Dan is using for SPEEDY), however as the current circulant embedding approach we are using only works for isotropic / stationary covariance functions anyway, I decided to keep things simple here and defer these changes to a separate PR.

test/optimal_filter_validation.yaml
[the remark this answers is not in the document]
giordano: > we can just use a simple exponential expression rather than having to use `SpecialFunctions` to load the `gamma` and `besselk` functions.

If the worry is to load another package during the tests, `SpecialFunctions` is already pulled in by `GaussianRandomFields`, so it isn't much more work.

> It might still be better to do this though I guess from a robustness standpoint so that the test is valid for any value of `nu`.

I think that'd be better, yes.

> Or at least I should perhaps put a comment in the YAML file to make clear this value shouldn't be changed?

At very least this, yes, please.

giordano (approved):
Looks great to me, after deciding what value of `nu` to use in the tests

test/optimal_filter_validation.jl
```diff
@@ -98,7 +98,7 @@ Sobs = 0.01
 #### Example Specification of Covariance Function
 function R(x::Float64, y::Float64, th)
 
-  Cov = (th.s^2)*exp(-(abs(x)+abs(y))/(2.0*th.l))
+  Cov = (th.s^2)*exp(-hypot(x, y) / th.l)
```
tkoskela, line 101: These two expressions don't look equal, was this an intentional correction?

src/OptimalFilter.jl
```diff
@@ -106,24 +111,25 @@ function covariance_grid_stations!(cov::AbstractMatrix{T}, grid::Grid, stations:
...
 
 end
 
-# Covariance between observations R_22, from equationd 3-4 in in Dietrich & Newsam 96
-# TODO: Ask Alex why we add sigma^2 on the diagonal
-function covariance_stations!(cov::AbstractMatrix{T}, grid::Grid, stations::NamedTuple, noise_params::NamedTuple, std::T) where T
+# Covariance between observations R̅₂₂ = R₂₂ + Σ, where R₂₂ is as defined in equations 3 and 4 in
+# Dietrich & Newsam 96 and Σ is the observation noise covariance (see note at end of
+# Section 2 in Dietrich & Newsam 1996)
+function covariance_stations!(cov::AbstractMatrix{T}, grid::Grid, stations::NamedTuple, covariance_structure::IsotropicCovarianceStructure{T}, std::T) where T
```
tkoskela, line 122: Thanks for catching and fixing the TODO 😄

test/optimal_filter_validation.jl
giordano: See discussion above https://github.com/Team-RADDISH/ParticleDA.jl/pull/216#discussion_r793748017 :wink:

tkoskela: Ah! I didn't notice the resolved discussion

tkoskela (approved):

Project.toml
```diff
@@ -9,11 +9,13 @@ EllipsisNotation = "da5c29d0-fa7d-589e-88eb-ea29b0a81949"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 FieldMetadata = "bf96fef3-21d2-5d20-8afa-0e7d4c32a885"
 Future = "9fa8497b-333b-5362-9e8d-4d0656e87820"
+GaussianRandomFields = "e4b2fa32-6e09-5554-b718-106ed5adafe9"
 HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
```
giordano, line 18: This isn't used here, is it?

matt-graham: Yeah sorry I just noticed this was incorrectly added; now removed. I think I added this when I initially got a missing dependency error when running the tests before realising I was using the wrong environment.

matt-graham:
I've now put the full definition of the Matern covariance function in `optimal_filter_validation.jl` so that the corresponding test should work for any `nu` value used in the YAML file and changed back the `nu` value used in the YAML file to the previous value of 2.5.

giordano:
Looks perfect!  Thanks a lot again!

Issue #215, opened 2022-01-21
Generalising the optimal proposal implementation


matt-graham:
As discussed in #208 the current implementation of sampling particles from the'optimal' proposal distribution makes a number of assumptions about the model that inference is being performed for - that it is defined on a regular rectangular 2D spatial grid, that the state noise is a Gaussian random field with stationary covariance, and that the observations are of a single scalar state variable at locations corresponding to grid points - that mean it cannot currently be applied to models such as SPEEDY for which one or more of these assumptions to do not apply. There are also various parts of the code which overlap with functionality separately implemented in the `GaussianRandomFields` package - one is the definition of the covariance function as discussed in #213 but also the circulant embedding approach for sampling from the optimal proposal (following the method described in [Dietrich and Newsam (1996)](https://doi.org/10.1029/94WR02977)) shares a lot of details with the circulant embedding approach implemented by  `GaussianRandomFields` which is used to sample (unconditioned) Gaussian random realizations to use as state noise. 

I *think* that by using a different approach for generating samples from the optimal proposal distribution we can generalize the optimal proposal to work for more general observation regimes and covariance structures of the state noise (including allowing arbitrary spatial dimensions, non-stationary covariance functions, non-structured grids, off-grid observations, observations of multiple state variables) while still retaining the ability to leverage efficient implementations for generating samples where extra structure is available (but without needing to for example reimplement functionality already provided by `GaussianRandomFields`).

Before describing the idea I'll first set up some notation. We assume a state-space model formulation where given a current state vector `x`, random vectors `X` and `Y` representing respectively the state and observations at the next time step are generated according to `X = F(x) + U` with `U ~ Normal(0, Q)` and  `Y = HX + V`   with `V ~ Normal(0, R)`. Under this model given observations `y` of `Y`, the optimal (in the sense of minimising the variance of the resulting importance weights) proposal distribution in a particle filter for the state is `Normal(m, C)` with `m = F(x) + QHᵀ(HQHᵀ + R)⁻¹(y − HF(x))` and `C = Q − QHᵀ(HQHᵀ + R)⁻¹HQ` - see for example [Doucet et al. (2000)](https://doi.org/10.1023/A:1008935410038) or [Snyder (2012)](https://www.ecmwf.int/en/elibrary/12354-particle-filters-optimal-proposal-and-high-dimensional-systems).

In our case specifically 
[...]

Issue #208, opened 2022-01-17
Generalise update_particle_proposal!


DanGiles:
update_particle_proposal! should be generalised to operate on different subarrays of the particles state vector. 

```Julia

function update_particle_proposal!(model_data, filter_data, filter_params, truth_observations, nprt_per_rank, filter_type::OptimalFilter)

        # Optimal Filter: After updating the particle dynamics, we apply the "optimal proposal" in
        #                 sample_height_proposal!() to the first state variable (height). We apply
        #                 a sample from the gaussian random field in update_particle_noise!() to the other
        #                 state variables (velocity).

        particles = get_particles(model_data)

        # Apply optimal proposal, the result will be in offline_matrices.samples
        sample_height_proposal!(@view(particles[:,:,1,:]),
                                filter_data.offline_matrices,
                                filter_data.online_matrices,
                                truth_observations,
                                filter_data.stations,
                                filter_data.grid,
                                filter_data.grid_ext,
                                filter_data.fft_plan,
                                filter_data.fft_plan!,
                                filter_params,
                                filter_data.rng[threadid()],
                                filter_data.obs_noise_std)

        # Add noise from the standard gaussian random field to all state variables in model_data
        update_particle_noise!(model_data, nprt_per_rank)

        # Overwrite the height state variable with the samples of the optimal proposal
        particles[:,:,1,:] .= filter_data.online_matrices.samples

```
For example, assimilating the surface pressure in the SPEEDY implementation requires:

```Julia
particles[:,:,1,5,:]
```

Perhaps one could define the specified indices of interest in filter params?

tkoskela:
Your example is a bit confusing because you've added a dimension, did you mean you want to have
```
particles[:,:,5,:]
```
I think your suggestion of adding a parameter for the index is a good one.

DanGiles:
Yes, we have added a dimension to the particle StateVectors in the SPEEDY implementation:

https://github.com/DanGiles/ParticleDA.jl/blob/733796f90e082fea9c2bd3dc4fb711e85bc00182/speedy/model/model.jl#L322


The variables are: 
https://github.com/DanGiles/ParticleDA.jl/blob/733796f90e082fea9c2bd3dc4fb711e85bc00182/speedy/model/model.jl#L191

As we are assimilating just the surface pressure (ground level in elevation), we want to access the subarray:

```
particles[:,:,1,5,:]
```
in the update_particle_proposal!

tkoskela:
[...]

Issue #205, opened 2021-11-25
Role of `weight_std` parameter not clear


matt-graham:
In the definition of the `FilterParameters` struct the `weight_std` parameter is documented as 

https://github.com/Team-RADDISH/ParticleDA.jl/blob/c130ec02a2d7e922ed0fc4f5c61e1c6f3c6dda60/src/params.jl#L15

however from how it is actually used in `ParticleDA.jl` where it is passed to this function

https://github.com/Team-RADDISH/ParticleDA.jl/blob/c130ec02a2d7e922ed0fc4f5c61e1c6f3c6dda60/src/ParticleDA.jl#L165-L179

it appears that it is actually the assumed standard deviation of the (isotropic Gaussian) observation noise.

The particle weights are random variables and do have a standard deviation, however in general this will not be equal to the observation noise standard deviation. The standard deviation of the weights can be a useful quantity to measure - for example it is related to [the effective sample size of a given set of weighted particles ](https://en.wikipedia.org/wiki/Effective_sample_size#Weighted_samples) - and so I think having a parameter named `weight_std` could lead to confusion about whether this is what is being represented by this value.

There is also a separate `obs_noise_std` parameter in the `ModelParameters` struct. This seems a more natural home for this value to me than `FilterParameters` as the observation process (including measurement noise) is part of the modelling assumptions rather than a parameter of the filter / inference method. I would say therefore it would be better to just remove `weight_std`  from `FilterParameters` and use the `obs_noise_std` attribute of `ModelParameters` in place, as most (all?) of the time these should have the same value. 

While there are possibly some use cases in a simulated data setting of using a different observation noise standard deviation to generate the observations compared to that assumed in the model used for inference (for example to test robustness to model mismatch) I think this would be better handled by having different `ModelParameters` instances for the 'true' underlying generative model (which in a real data assimilation setting is unknown) and the model being inferred from the observations using the filter.

PR #218, opened 2022-03-01
Generalising the optimal proposal implementation

DanGiles:
This is a draft pull request for a branch which contains the edits outlined by Matt in #215


src/ParticleDA.jl
```diff
@@ -384,35 +399,9 @@ function update_particle_proposal!(model_data, filter_data, truth_observations,
...
-
-        # Optimal Filter: After updating the particle dynamics, we apply the "optimal proposal" in
-        #                 sample_height_proposal!() to the first state variable (height). We apply
-        #                 a sample from the gaussian random field in update_particle_noise!() to the other
-        #                 state variables (velocity).
-
-        particles = get_particles(model_data)
-
-        # Apply optimal proposal, the result will be in offline_matrices.samples
-        sample_height_proposal!(@view(particles[:,:,1,:]),
```
giordano, line 396: Quick note that `sample_height_proposal!` is unused, so it may be removed completely (also because as noted in https://github.com/Team-RADDISH/ParticleDA.jl/issues/208#issuecomment-1017245537 the name was too specific to the TDAC model)

src/ParticleDA.jl
```diff
@@ -384,35 +399,9 @@ function update_particle_proposal!(model_data, filter_data, truth_observations,
...
-
-        # Add noise from the standard gaussian random field to all state variables in model_data
-        update_particle_noise!(model_data, nprt_per_rank)
-
-        # Overwrite the height state variable with the samples of the optimal proposal
-        particles[:,:,1,:] .= filter_data.online_matrices.samples
-        set_particles!(model_data, particles)
-
+    update_particle_noise!(model_data, nprt_per_rank)
+    indices = get_indices(model_data)
```
giordano, line 403: `indices` is unused?

test/model/model.jl
```diff
@@ -485,6 +488,19 @@ function ParticleDA.update_truth!(d::ModelData, _)
...
 
+function ParticleDA.sample_observations_given_particles(d::ModelData, nprt_per_rank)
+    
+    for ip in 1:nprt_per_rank
+        get_obs!(@view(d.observations.model[:,ip]),
+                 @view(d.states.particles[:, :, :, ip]),
+                 d.stations.ist,
+                 d.stations.jst,
+                 d.model_params)
+        add_noise!(d.observations.model[:,ip], d.rng[1], d.model_params)
```
giordano, line 499: Note that slicing creates a copy, so here you aren't actually modifying `d.observations.model` in-place.  You need to use [`@view`](https://docs.julialang.org/en/v1/base/arrays/#Base.@view) to take a view of the array you're indexing, as done above.  See for example
```julia
julia> f!(vec) = vec .= 1.0
f! (generic function with 1 method)

julia> vec = randn(3)
3-element Vector{Float64}:
 1.0176810348984888
 1.2137544102635487
 0.6981311184850002

julia> f!(vec[:])
3-element Vector{Float64}:
 1.0
 1.0
 1.0

julia> vec # Oops... the previous function call didn't actually modify `vec`!
3-element Vector{Float64}:
 1.0176810348984888
 1.2137544102635487
 0.6981311184850002

julia> f!(@view(vec[:]))
3-element view(::Vector{Float64}, :) with eltype Float64:
 1.0
 1.0
 1.0

julia> vec # This one worked as expected
3-element Vector{Float64}:
 1.0
 1.0
 1.0
```

matt-graham: In addition to @giordano's comment I think `d.rng[1]` here should also be `d.rng[threadid()]` as we want the thread-specific random number generator.

src/ParticleDA.jl
```diff
@@ -42,6 +42,12 @@ Return the number of state variables.
 """
 function get_n_state_var end
 
+"""
+    ParticleDA.get_indices(model_data) -> Vector{Int}
+
+Return the vector containing the indices of assimilated values in the state vector
+"""
+function get_indices end
```
matt-graham, line 50: Is the intention here is for this to specifically return the indices along only the 'state variable' axis in to the state array as the method definition for the TDAC model current returns `[1]` which I guess would correspond to the height component? Structuring like this would I think limit us to having the same set of state variables observed at all observation locations / stations (as it seems like this would be specifying the overall set of observation indices as a Cartesian product between the observed state variable indices and indices corresponding to each observation location). While it's probably reasonable to only provide an implementation for that case for now until we need the extra generality, from the perspective of generalising the optimal filter implementation I would say there is a benefit to trying to avoid depending on the state particles being of a particular spatial dimension / shape, and just interfacing with each particle as a flat vector. If `get_indices` instead was specified returned a vector of integer indices in to the flattened states corresponding to observed components that would allow for this abstraction, while still allowing the definitions in particular models to assume that the state variables observed at each location are the same. It might also be worth giving this a more informative name like `get_observed_state_indices`.

test/model/model.jl
giordano: I've been thinking we may want move to the new task-reproducible RNG introduced in v1.7: https://julialang.org/blog/2021/11/julia-1.7-highlights/  But it isn't clear to me how that'd work with StableRNGs we use in the tests.

src/ParticleDA.jl
DanGiles: Yes, this has been removed. At the time I was thinking of just passing the observed state variables to update_particles_given_observations!

test/model/model.jl
DanGiles: Thanks for explaining the difference. I have updated as suggested. @giordano let me know when you have decided on the task-reproducible RNG.

giordano: I've just realised that we're using the reference data with StableRNGs for single-thread programs anyway (I guess because we know that RNG isn't reliable in multi-threaded situations), so switching to that might not be bad at all. I'll try to have a look next week, this will mean we'll have to require Julia v1.7.

src/ParticleDA.jl
DanGiles: As suggested I have renamed it to be get_observed_state_indices. I don't see it being too difficult to increase the generality to interface with flattened states. It helps that it is defined on the model side.

matt-graham: Another reason for trying to keep the interface in terms of flattened vectors that occurred to me in the meeting yesterday is that this would also make it easier to generalise the particle filter implementation (both bootstrap and optimal proposals) to deal with non-spatial models.

DanGiles:
Latest version is a working example of the generalised optimal filter. I have tested it with both TDAC and speedy. Visually the results with TDAC are comparable to the old optimal filter. 

- Do you have a suggested for the best way of defining the cholesky type in [OfflineMatrices](https://github.com/Team-RADDISH/ParticleDA.jl/blob/5ead4a2e3c834388553e8eb879b66eaa3121a33b/src/OptimalFilter.jl#L238)? The current implementation might not be optimal.

matt-graham:
> Do you have a suggested for the best way of defining the cholesky type in [OfflineMatrices](https://github.com/Team-RADDISH/ParticleDA.jl/blob/5ead4a2e3c834388553e8eb879b66eaa3121a33b/src/OptimalFilter.jl#L238)? The current implementation might not be optimal.

The type returned by `cholesky` is `Cholesky` so I think something like `Cholesky{F, Matrix{F}}` should work?

matt-graham:
> * Do you have a suggested for the best way of defining the cholesky type in [OfflineMatrices](https://github.com/Team-RADDISH/ParticleDA.jl/blob/5ead4a2e3c834388553e8eb879b66eaa3121a33b/src/OptimalFilter.jl#L238)? The current implementation might not be optimal.

Having looked at this in a bit more detail now - we need to both:

(i) Add a new type parameter to the `OfflineMatrices` struct specifically for the `chol_covariances_observations` field in 

https://github.com/Team-RADDISH/ParticleDA.jl/blob/5ead4a2e3c834388553e8eb879b66eaa3121a33b/src/OptimalFilter.jl#L8-L25

as this will be of a different concrete type `Cholesky{...}` to the other fields with the same type parameter `T` in the current implementation, this should be something like

```Julia
 struct OfflineMatrices{T<:AbstractArray, S<:AbstractArray, U<:AbstractArray, V<:Cholesky} 
     rho_bar::S 
     R12::T 
     R21_bar::T 
     R22::T # Covariance between observation stations 
     R22_inv::T # Inverse of R22 
     chol_covariance_observations::V 
     R12_invR22::T 
     Lambda::S 
     K::U 
     L::T 
     mu20::T 
     buf1::T 
     buf2::T 
end  
```

Alternatively I think we could have something like the following

```Julia
struct OfflineMatrices{
    R<:Real, 
    C<:Complex{R}, 
    RV<:AbstractVector{R}, 
    RM<:AbstractMatrix{R}, 
    CM<:AbstractMatrix{C}, 
    RC<:Cholesky{R, RM}
} 
     rho_bar::RV 
     R12::RM
     R21_bar::RM 
     R22::RM # Covariance between observation stations 
     R22_inv::RM # Inverse of R22 
     chol_covariance_observations::RC 
     R12_invR22::RM 
     Lambda::RV 
     K::CM 
     L::RM 
     mu20::RM 
     buf1::RM 
     buf2::RM
end  
```
which to me would make it clearer what the role of the different parametric types is and put tighter requirements on the allowable types, though not sure if there are disadvantages to declaring the struct like this @giordano? 

(ii) Change how this field is initialised in `init_offline_matrices`.

For this I think ideally we would leave the field undefined when the struct instances is first created in 

https://github.com/Team-RADDISH/ParticleDA.jl/blob/5ead4a2e3c834388553e8eb879b66eaa3121a33b/src/OptimalFilter.jl#L233-L246

however I don't think that is trivial for the `Cholesky` type. While we can do something like

```Julia
Cholesky(Matrix{F}(undef, stations.nst, stations.nst), 'l', 0)
```

it's not clear that we can later update this `Cholesky` object in place with the actual Cholesky factor without resorting to calling [an internal method `_chol!`](https://github.com/JuliaLang/julia/blob/d29d2d622ae416c3e5ba82cb55ef0535a75fe8e8/stdlib/LinearAlgebra/src/cholesky.jl#L188-L261) and even then it is still unclear if it would be possible to recycle the memory allocated `Matrix{F}(undef, stations.nst, stations.nst)`. 

I think it would therefore probably make more sense to switch from the current pattern of first initialising the `OfflineMatrices` struct with undefined arrays and then computing the values in-place to instead first computing the values and then creating the struct from the already assigned arrays. As we also can also remove a lot of the entries in `OfflineMatrices` as these are no longer required with the new implementation, this shouldn't be too difficult to refactor.

giordano:
> which to me would make it clearer what the role of the different parametric types is and put tighter requirements on the allowable types, though not sure if there are disadvantages to declaring the struct like this @giordano?

No, I don't expect any particular problems with that.

And I agree that peeking into internals like `_chol!` is to be avoided as much as possible, your proposal looks reasonable to me (but I'm less familiar with the linear algebra internals, so I can't think of all possible consequences right now)

src/ParticleDA.jl
```diff
@@ -180,35 +180,45 @@ of [`run_particle_filter`](@ref) to select the bootstrap filter.
...
-    weight .= Distributions.logpdf(Distributions.MvNormal(obs, cov_obs), obs_model)
-
+function compute_individual_particle_log_weight(
+    observations::AbstractVector{T},
+    observations_mean::AbstractVector{T},
+    filter_data::NamedTuple,
+    ::BootstrapFilter,
+) where T
+    difference = observations - observations_mean
+    return -0.5 * sum(difference.^2) / filter_data.obs_noise_std^2
```
DanGiles, line 210: We have discussed this before but should we retain the difference between obs_noise_std and the previously defined filter_params.weight_std? Perhaps with more suitable naming?

DanGiles:
I ran this version with TDAC (same parameter set up as before). The minimum value for the std of the assimilated wave height no longer coincides with the observation locations, any ideas on why?
TDAC

src/ParticleDA.jl
matt-graham: Ah good point yes we should probably sort this as part of this PR. I'd vote for just getting rid of `filter_params.weight_std` under the rationale in #205.

DanGiles:
Yes you're right @giordano , my bad!

src/ParticleDA.jl
DanGiles: I agree, I have just updated the PR with the removal of weight_std

giordano:
No problem, I remember it because I've been hit by this mix up a few times myself :slightly_smiling_face:  I believe this has to do with the convention used by the plotting package of where the origin coordinate is (typically top or bottom left).

test/model/model.jl
```diff
@@ -442,7 +445,7 @@ ParticleDA.get_obs_noise_std(d::ModelData) = d.model_params.obs_noise_std
 ParticleDA.get_model_noise_params(d::ModelData) = Matern(d.model_params.lambda[1],
                                                          d.model_params.nu[1],
                                                          σ=d.model_params.sigma[1])
```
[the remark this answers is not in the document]
matt-graham: I think this all of the values are used in `init_gaussian_random_field_generator` to create one generator per state variable. The usage above is as we're currently only observing the first state variable and so only need the optimal filter to be able to access the covariance structure for this variable

matt-graham:
The `Optimal Filter unit tests` test set have now been updated to test all the functions in the update `OptimalFilter.jl` file. 

Currently the `Optimal Filter validation` test set is still failing as it is currently using bits of the old implementation that are no longer available. While we could update to doing something similar with the new implementation the script from Alex B we are validating against is using the previous approach from the Dietrich and Newsam (1996) paper, so we will get different results across the two implementations (in terms of the sampled values, distributionally they should be the same). @tkoskela @giordano should I just remove the validation test set?

matt-graham:
As an alternative to the current validation we could do a more general test that for a large number of particles the empirical distribution of the updated particles given the previous particles is consistent with the expected normal distribution `Normal(m, C)` with `m = F(x) + QHᵀ(HQHᵀ + R)⁻¹(y − HF(x))` and `C = Q − QHᵀ(HQHᵀ + R)⁻¹HQ` (notation as in #215). This would just require sampling from the proposal for a relatively large number of particles and checking the empirical mean and covariance are consistent with the analytic values with some tolerance.

matt-graham:
I think this is now ready for review 🎉!

I have updated the unit tests for the optimal proposal and added a new validation test that checks that the empirical moments of the updated particles have errors from the expected true normal distribution parameters that scale with the expected `O(1/√N)` Monte Carlo error.

As part of this I discovered that previously the `add_noise!` method in `model.jl` which adds observation noise to the model observations, was previously sampling from a truncated normal distribution
https://github.com/Team-RADDISH/ParticleDA.jl/blob/de4336a0d2eb4f10307205aa9adce91b378d7646/test/model/model.jl#L326-L332
The optimal proposal implementation we are using requires the observation noise to be Gaussian and the comment before the function also implies the observation noise is intended to be Gaussian. Comfortingly the validation test was failing before this change as the empirical moments did not match the expected values but passed after this fix so it does seem to be sensitive enough to pick up on such bugs.

giordano (approved):
Looks good to me, great work Dan and Matt!

Incidentally, the performance of the small MPI benchmark with the optimal filter went from
```
                                        Time                    Allocations      
                               ───────────────────────   ────────────────────────
       Tot / % measured:            17.3s /  99.9%           1.28GiB / 100.0%    
```
to
```
                                        Time                    Allocations      
                               ───────────────────────   ────────────────────────
       Tot / % measured:            5.84s /  99.7%            362MiB / 100.0%    
```
In particular the filter initialisation is now much better: [before](https://github.com/Team-RADDISH/ParticleDA.jl/runs/6127604216?check_suite_focus=true#step:6:223)
```
 Section               ncalls     time    %tot     avg     alloc    %tot      avg
 ────────────────────────────────────────────────────────────────────────────────
 Filter initialization      1    10.4s   60.3%   10.4s   0.96GiB   75.4%  0.96GiB
 Particle Proposal          5    3.68s   21.3%   736ms    320MiB   24.5%  64.0MiB
```
and [after](https://github.com/Team-RADDISH/ParticleDA.jl/runs/6273622653?check_suite_focus=true#step:6:223):
```
 Section               ncalls     time    %tot     avg     alloc    %tot      avg
 ────────────────────────────────────────────────────────────────────────────────
 Particle Proposal          5    2.46s   42.2%   491ms    192MiB   52.9%  38.3MiB
 Filter initialization      1    1.67s   28.6%   1.67s    147MiB   40.7%   147MiB
```

benchmark/benchmarks.jl
```diff
@@ -47,7 +47,15 @@ SUITE["base"]["get_mean_and_var!"] = @benchmarkable ParticleDA.get_mean_and_var!
...
 SUITE["base"]["update_particle_dynamics!"] = @benchmarkable ParticleDA.update_particle_dynamics!($(model_data), $(nprt_per_rank))
 SUITE["base"]["update_particle_noise!"] = @benchmarkable ParticleDA.update_particle_noise!($(model_data), $(nprt_per_rank))
-SUITE["base"]["get_log_weights!"] = @benchmarkable ParticleDA.get_log_weights!(weights, truth_observations, model_observations, $(filter_params.weight_std)) setup=(weights = Vector{Float64}(undef, nprt_per_rank); truth_observations=ParticleDA.update_truth!(model_data, nprt_per_rank); model_observations = ParticleDA.get_particle_observations!(model_data, nprt_per_rank))
+SUITE["base"]["get_log_weights!"] = @benchmarkable ParticleDA.get_log_weights!(
+    weights, truth_observations,  model_observations, filter_data, filter_type
+) setup=(
+    weights = Vector{Float64}(undef, nprt_per_rank); 
+    truth_observations=ParticleDA.update_truth!(model_data, nprt_per_rank); 
+    model_observations = ParticleDA.get_particle_observations!(model_data, nprt_per_rank);
+    filter_data = (obs_noise_std=bootstrap_filter_data.obs_noise_std,);
```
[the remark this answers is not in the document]
matt-graham: Ah thanks yeah that is a useful pattern to know about, I struggled to spot initially that it was a missing trailing comma that was causing the benchmark check to fail!

tkoskela (approved):
I had a final look through the changes and it looks good to me. Great work Matt, Dan and Mose! I made some comments about things that were not entirely clear in the tests

test/runtests.jl
```diff
@@ -289,120 +283,183 @@ end
...
+        # update_particle_dynamics should be deterministic so all particles remain equal 
+        # to each other
+        @test all(firstonlastaxis(particles) .== particles)
+        observations_mean_given_particle = copy(firstonlastaxis(
+            ParticleDA.get_particle_observations!(model_data, nprt_per_rank)
+        ))
+        ParticleDA.update_particle_noise!(model_data, nprt_per_rank)
+        noised_particles = copy(ParticleDA.get_particles(model_data))
+        # update_particle_noise! should add noise to all particle components
+        @test all(particles .!= noised_particles)
```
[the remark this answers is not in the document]
matt-graham: I guess that technically because of limited floating point-precision there is a non-zero probability of one or more noise components being exactly zero and for a large enough number of dimensions it could become non-negligible, though I think it would need to be very high: even for `Float16` precision it seems to take ~10⁷ draws from `randn` on average to get an exactly zero value (and my intuition would be for `Float64` it will take roughly 2⁴⁸ times as many draws on average).

Given the state noise covariance is being computed explicitly I can make this test more specific to check that the noise is approximately zero mean and of correct covariance to within Monte Carlo error.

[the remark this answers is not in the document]
matt-graham: Now removed in favour of checking state noise (of observed state components) has correct covariance.

test/runtests.jl
```diff
@@ -289,120 +283,183 @@ end
...
+        # noise added by update_particle_noise! should be zero mean and so mean of
+        # noised particles should be within O(1/√N) Monte Carlo error of particles
+        @test maximum(
+            abs.(
+                mean(noised_particles, dims=ndims(noised_particles)) 
+                .- firstonlastaxis(particles)
+            )
+        # 4 constant in here has been based on looking at scale of typical deviations -
+        # main point of check is that errors scale at expected O(1/√N) rate
+        ) < (4. / sqrt(nprt))
```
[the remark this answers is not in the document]
matt-graham: Now shifted these comments to be before the tests in all cases

PR #219, opened 2022-03-14
Require Julia v1.7 and use new task-local RNG

giordano:
In Julia v1.7 [has been introduced](https://julialang.org/blog/2021/11/julia-1.7-highlights/#new_rng_reproducible_rng_in_tasks) a new random number generator (RNG) which is task-reproducible.  For example, take this function which sets all the elements of an array to random numbers with normal distribution and then returns the sum of all the elements:
```julia
using Random
function test!(a::AbstractArray, rng::Random.AbstractRNG)
    Threads.@threads for idx in eachindex(a)
        a[idx] = randn(rng)
    end
    return sum(a)
end
```
If you start julia v1.7 or following versions with, say, 4 threads (either export the environment variable `JULIA_NUM_THREADS=4` or use `julia -t 4`), then you can verify that calling this function 1 million times with the same RNG seed you always get the same result:
```julia
julia> VERSION < v"1.8.0-" && (allequal(itr) = all(isequal(first(itr)), itr));

julia> allequal(test!(Vector{Float64}(undef, 1000), Random.seed!(42)) for _ in 1:1_000_000)
true

julia> Base.Threads.nthreads()
4

julia> VERSION
v"1.7.2"
```
This doesn't happen if you use julia v1.6-:
```julia
julia> VERSION < v"1.8.0-" && (allequal(itr) = all(isequal(first(itr)), itr));

julia> allequal(test!(Vector{Float64}(undef, 1000), Random.seed!(42)) for _ in 1:1_000_000)
false

julia> Base.Threads.nthreads()
4

julia> VERSION
v"1.6.4"
```
because the old RNG isn't reproducible when using multiple threads, so we had to use annoying tricks like generating a _long_ vector of RNGs and accessing it by `threadid()`.

Not only is the new RNG task-reproducible, but it's also actually faster when using multiple threads, because it automatically accesses a task-local RNG, instead of a global shared one, but this wasn't a concern for us as we using the annoying trick above.

Ref: https://github.com/Team-RADDISH/ParticleDA.jl/pull/218#discussion_r822815275


matt-graham (approved):
This seems pretty neat 🙂. As far as I can see you have changed all the occurrences where a vector of RNGs was previously being used. I am not sure what is causing the failing benchmark job - running the benchmarks locally from this branch works fine and the error in the logs `LoadError: MethodError: no method matching randjump(::TaskLocalRNG, ::BigInt)` and traceback is not particularly revealing (to me at least), particularly as there do not seem to be any `randjump` calls in our code at least after the changes in the branch.

Just to check I am understanding what is going on with the task-local RNGs correctly - when running multithreaded each task is assigned a local RNG which is seeded from the state of the RNG of the parent task and in our case we are use threading to parallelise updates over the particles associated with each MPI rank. So a task will involve updates to a single particle and we will have a RNG per particle on each MPI rank? How are the task-local RNGs across different MPI ranks related?

giordano:
> This seems pretty neat 🙂. As far as I can see you have changed all the occurrences where a vector of RNGs was previously being used.

Correct, because before we were generating a looong vector to simulate having different RNGs per threads.  By using the default thread-safe RNG we don't need to do that.  The caveat is that now this assumes we either use a single thread (what we do in the tests when we use StableRNG) or the RNG must be thread-safe.  Is that acceptable?  Probably this should be documented explicitly, although I don't really expect users to mess up with the RNG, we added the argument mainly for easier testing.

> I am not sure what is causing the failing benchmark job - running the benchmarks locally from this branch works fine and the error in the logs `LoadError: MethodError: no method matching randjump(::TaskLocalRNG, ::BigInt)` and traceback is not particularly revealing (to me at least), particularly as there do not seem to be any `randjump` calls in our code at least after the changes in the branch.

Yeah, I should have a closer look at the benchmarks failure.  I can't find anything in the package directory if I grep `randjump`, I suspect this happens when the benchmark script in this PR is run on `master`.

> Just to check I am understanding what is going on with the task-local RNGs correctly - when running multithreaded each task is assigned a local RNG which is seeded from the state of the RNG of the parent task and in our case we are use threading to parallelise updates over the particles associated with each MPI rank. So a task will involve updates to a single particle and we will have a RNG per particle on each MPI rank?

That sounds about right!

> How are the task-local RNGs across different MPI ranks related?

That's a great question to which I don't have a definitive answer!  I presume they'll be basically be seeded independently in each process, but if we fixed the seed we should get the same stream of numbers for each thread in each MPI process.

> Out of interest is `Random.default_rng` documented anywhere? I couldn't find anything at https://docs.julialang.org/en/v1.7/stdlib/Random/

Ugh, the perils of looking too much into the source code!  I've always used that when I wanted to reference the default RNG, I probably had checked what `rand(Float64)` does (`@less rand(Float64)`), but you're right this isn't documented.  I'm investigating to find out what's the "official" way to call it.

> Bad experiences with [mutable default arguments in Python](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) have made me wary about this sort of usage but from what I can tell from how optional arguments are implemented in Julia it seems this might not be a worry here?

Correct, Julia doesn't do that: https://twitter.com/MoseGiordano/status/1450564075604398081.  Default arguments are evaluated at call-time, this is also mentioned in the list of [noteworthy differences from Python](https://docs.julialang.org/en/v1/manual/noteworthy-differences/#Noteworthy-differences-from-Python), where this example is proposed:
```julia
julia> f(x=randn()) = x
f (generic function with 2 methods)

julia> f()
1.0377860667759702

julia> f()
-1.1612721874205534

julia> f()
0.8992181216982671

julia> f()
-0.2261044092292596
```

matt-graham:
> Correct, Julia doesn't do that: https://twitter.com/MoseGiordano/status/1450564075604398081. Default arguments are evaluated at call-time, this is also mentioned in the list of [noteworthy differences from Python](https://docs.julialang.org/en/v1/manual/noteworthy-differences/#Noteworthy-differences-from-Python), where this example is proposed:

Ah thanks, I thought I'd read this somewhere but couldn't see where in the docs this was 👍

> That's a great question to which I don't have a definitive answer! I presume they'll be basically be seeded independently in each process, but if we fixed the seed we should get the same stream of numbers for each thread in each MPI process.

If we do get the same stream of random number in each thread / task across different MPI ranks when seeding this could cause issues with the particles sets across different ranks being coupled (not sure if this is what you meant or that we would get different sets of RNG states across ranks but reproducibly). If we use the same task-local RNGs to also set the initial states such that we get particles with the same set of initial states on each rank then this might end up with the set of particle trajectories across all MPI ranks being the same I think? Even if they start from different initial states, if the dynamics of the model tend to drive the particles together it could be they will begin to coalesce over time if using common streams of random variates (this is sometimes purposefully done as a variance reduction approach in some Monte Carlo algorithms). This might not be a problem, but probably worth checking!

giordano:
I've looked into the benchmark failure: I'm moderately sure that's due to the fact the API changed between `master` and this PR.  My understanding is that, in order to do a consistent benchmark between the reference branch and the current branch, [`BenchmarkCI.jl`](https://github.com/tkf/BenchmarkCI.jl) runs the same script on both branches, which breaks in our case.  I'm moderately sure we've had similar problems in previous PRs which similarly changed the API.  Side note, benchmarking in CI has been quite unreliable in the past few months, it was much better when we initially set it up, but I never got the time to investigate that further.

Regarding the default RNG to use here, I changed to `TaskLocalRNG()`.  I think that's better than using `default_rng` because

1. it's [documented](https://docs.julialang.org/en/v1/stdlib/Random/#Random.TaskLocalRNG), so the API should be more stable for the whole v1.x series of Julia
2. it better describes what we actually want: a task local RNG, whether it's the default or not.

Also, I went through [the documentation](https://team-raddish.github.io/ParticleDA.jl/dev/), it doesn't look like we document the RNG anywhere, only how to pass a fixed seed, which pretty much reflects what I said above: we don't expect/want users to mess up with the RNG themselves.

Regarding MPI, I did some tests, I'm now more convinced that in each MPI process the task-local RNGs are seeded independently, unless of course you fix the seed yourself.  For example, using this script:
```julia
using Base.Threads, MPI, Random

MPI.Init()

function test(rng::Random.AbstractRNG)
    a = Vector{Tuple{Float64,Int}}(undef, 2 * nthreads())
    Threads.@threads for idx in eachindex(a)
        a[idx] = randn(rng), threadid()
    end
    return (a)
end

my_rank = MPI.Comm_rank(MPI.COMM_WORLD)
rng = Random.default_rng() # Random.seed!(1)
result = test(rng)
@info "result = $(result)" my_rank VERSION
```
I get
```console
% mpiexecjl -np 4 --project=/tmp julia --color=yes -t3 foo.jl
┌ Info: result = [(-1.268581937388697, 1), (0.5159211921034663, 1), (1.4989484564978073, 2), (-1.2261051396727385, 2), (-1.0654081052286541, 3), (0.7217060947370498, 3)]
│   my_rank = 2
└   VERSION = v"1.7.2"
┌ Info: result = [(0.0812626604290931, 1), (-0.6803388765264494, 1), (0.341276047759174, 2), (0.6468091567208047, 2), (-0.6443922557891207, 3), (-0.7087392371036914, 3)]
│   my_rank = 0
└   VERSION = v"1.7.2"
┌ Info: result = [(0.28024965204568075, 1), (-0.9433214116807365, 1), (-1.9771778793835806, 2), (0.561767676016203, 2), (0.033882308271889705, 3), (0.910944278609998, 3)]
│   my_rank = 3
└   VERSION = v"1.7.2"
┌ Info: result = [(1.0945598134825743, 1), (0.5709628099610049, 1), (-1.1482705096678063, 2), (-0.14921208742651323, 2), (1.0415112678943026, 3), (-1.4199587779115115, 3)]
│   my_rank = 1
└   VERSION = v"1.7.2"
```
If set `rng = Random.seed!(1)` I get the same results across all MPI ranks, but each thread still has an independent stream of numbers:
```julia
% ~/.julia/bin/mpiexecjl -np 4 --project=/tmp julia --color=yes -t3 foo.jl
┌ Info: result = [(-0.02799823154172939, 1), (1.483993806845476, 1), (-0.501685623035822, 2), (0.46986448243520346, 2), (0.2114792974372784, 3), (1.7709525255061302, 3)]
│   my_rank = 3
└   VERSION = v"1.7.2"
┌ Info: result = [(-0.02799823154172939, 1), (1.483993806845476, 1), (-0.501685623035822, 2), (0.46986448243520346, 2), (0.2114792974372784, 3), (1.7709525255061302, 3)]
│   my_rank = 2
└   VERSION = v"1.7.2"
┌ Info: result = [(-0.02799823154172939, 1), (1.483993806845476, 1), (-0.501685623035822, 2), (0.46986448243520346, 2), (0.2114792974372784, 3), (1.7709525255061302, 3)]
│   my_rank = 0
└   VERSION = v"1.7.2"
┌ Info: result = [(-0.02799823154172939, 1), (1.483993806845476, 1), (-0.501685623035822, 2), (0.46986448243520346, 2), (0.2114792974372784, 3), (1.7709525255061302, 3)]
│   my_rank = 1
└   VERSION = v"1.7.2"
```
Note that when running this last version of the script with Julia v1.6 I get
```console
% mpiexecjl -np 4 --project=/tmp julia-16 --color=yes -t3 foo.jl
┌ Info: result = [(0.2972879845354616, 1), (0.3823959677906078, 1), (-0.5976344767282311, 2), (-0.01044524463737564, 2), (-0.839026854388764, 3), (0.31111133849833383, 3)]
│   my_rank = 3
└   VERSION = v"1.6.4"
┌ Info: result = [(-1.0957880343275386, 1), (-0.2730708187609759, 1), (-1.437126476092288, 2), (0.08341763320751519, 2), (0.2972879845354616, 3), (0.3823959677906078, 3)]
│   my_rank = 2
└   VERSION = v"1.6.4"
┌ Info: result = [(-1.0957880343275386, 1), (-0.2730708187609759, 1), (0.2972879845354616, 2), (0.3823959677906078, 2), (-1.437126476092288, 3), (0.08341763320751519, 3)]
│   my_rank = 0
└   VERSION = v"1.6.4"
┌ Info: result = [(-0.839026854388764, 1), (0.31111133849833383, 1), (0.2972879845354616, 2), (0.3823959677906078, 2), (-0.5976344767282311, 3), (-0.01044524463737564, 3)]
│   my_rank = 1
└   VERSION = v"1.6.4"
```
The stream of numbers is not exactly the same, some numbers appear everywhere, but in different threads, while some other numbers appear only in some processes.

So I think that if we don't fix the seed manually (which we don't) we should be all good, even across multiple MPI processes.  Does it sound good?

matt-graham:
> The stream of numbers is not exactly the same, some numbers appear everywhere, but in different threads, while some other numbers appear only in some processes.
>
> So I think that if we don't fix the seed manually (which we don't) we should be all good, even across multiple MPI processes. Does it sound good?

I am not sure how typical I am, but I would pretty much always explicitly seed a RNG and pass it as an argument if possible whenever using a Monte Carlo algorithm like this, to make it possible to reproduce results / debug things that went wrong. So for example if I was running the test example in the documentation my default inclination would be to instead run something like the following

```Julia
...
rng = Random.seed!(6247203192634562690)
run_particle_filter(Model.init, input_file, BootstrapFilter(), rng=rng)
```

From what I can tell from the results of the tests in your comment this would mean that if I ran a script like this using MPI each MPI rank would run with the same top-level RNG state (and same set of spawned task-local RNGs) and so share the streams of random variates being produced? If this is the case, as this would potentially give incorrect results, I think we should probably make it clearer that the `rng` argument needs to be set carefully when running with MPI and/or provide an interface to allow reproducibility when using MPI without running in to this issue by explicitly constructing per-rank RNGs with different states from a single seed state. It seems this would equally apply to the current implementation with a vector of RNGs as it does to the thread-local RNGs here, so (providing I am not completely misunderstanding things!) this is probably worth creating a separate issue for.

giordano:
@matt-graham to be clear, are we now good to go with this PR?  I think #221 is somewhat orthogonal to this change, and this PR shouldn't be a regression with regard to that

matt-graham:
> @matt-graham to be clear, are we now good to go with this PR?  I think #221 is somewhat orthogonal to this change, and this PR shouldn't be a regression with regard to that

Yes agreed. Looks good to go to me.

PR #222, opened 2022-04-20
Collect code coverage and add badges to README file

giordano:


giordano:
Strictly speaking, it's an unrelated clean-up, but I thought that while I was there I could also update the caching wokflows :innocent:

PR #226, opened 2022-05-11
Assimilating multiple variables

DanGiles:
First attempt at assimilating multiple variables. Working example with TDAC and assimilating all the state variables: height, vx and vy. It is running but fails the various unit test due to changed function definitions, I will fix this once we have settled on some more important design choices. @matt-graham when you get the chance it would be great to discuss. I know you have a big deadline this week so no rush. 

Some immediate design points that come to mind are as follows:
- Should we have a vectorised obs_noise_std with a different value for each variable?
- What is the best way to combine the differences when calculating the particle weights? At present I have implemented a testing example where a [weighted sum](https://github.com/Team-RADDISH/ParticleDA.jl/blob/554fae8101bfdf330255bc7ed76ef38df1245c5a/src/ParticleDA.jl#L218-L231)  is calculated. If we are happy with this design, the relative_weights can be passed as model params
- Related to obs_noise_std should we scale the fact_cov_Y_Y differently for each variable?


matt-graham:
@DanGiles I think this is now superseded by the changes that were merged in #218 which allow assimilating multiple variables - shall we close this?

PR #232, opened 2022-08-22
Refactor: using flat vector states and decoupling states from model

### LINKED ISSUE #224: Observations currently generated within filtering loop
At the moment the observations conditioned on at each time step in the filtering loop are generated by calling `update_truth!` on each time step

https://github.com/Team-RADDISH/ParticleDA.jl/blob/093803153971862ff7d7bec9372ff8fa120501f8/src/ParticleDA.jl#L458-L461

In use cases with data from real-world systems the observations would be inputs to the filtering algorithm rather than being generated alongside the filtering. While the current design doesn't preclude having the observed data time series as a field in model data and having `update_truth!` just return an array corresponding to the current time step, the naming of the `update_truth!` would make this usage a bit confusing and this would require having a separate internal counter of the current time step in the `model_data` which is a bit brittle. 

Conceptually I also think it would cleaner to separate the observations and model - there might be several different models we want to calibrate against the same observed data, or we might want to test calibrating a model against data generated under a different model to check robustness to model misspecification. From a testing perspective ideally we would should also be able to run the filtering algorithm multiple times for the same observed data to check if we get filtering estimates that are distributionally consistent, and also for example to [check the calibration](https://arxiv.org/abs/2012.12670) of the distributional output (in the sense of whether the true states are plausible as samples from the estimated filtering distributions).

### LINKED ISSUE #229: Inconsistent conventions around spatial grid coordinates in LLW2d model
The default model parameters in `test/model/model.jl` specify a spatial grid with parameters

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/test/model/model.jl#L64-L69

The use of `nx - 1` and `ny - 1` in the denominators of `dx` and `dy` I think therefore suggests that the grid coordinates along the x-direction are `(0:(nx - 1)) * dx == [0, dx, 2 * dx, ..., (nx - 1) * dx]` and along the y-direction `(0:(ny - 1)) * dy == [0, dy, 2 * dy, ..., (ny - 1) * dy]`.  This is consistent with the outputs of this function to get the grid coordinates along the two axes

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/test/model/model.jl#L235-L241

The grid point corresponding to an index pair `(i, j) ∈ (1:nx) × (1:ny)` is therefore 
```
(x, y) = ((i - 1) * dx, (j - 1)  * dx)
```
The index pair of the grid point _nearest_ to a point `(x, y)` is 
```
(i, j) = (round(Int, x / dx) + 1, round(Int, x / dy) + 1)
```

When converting from station points to indices in the `set_stations!` methods, both for stations locations generated on a rectilinear grid

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/test/model/llw2d.jl#L34-L38

and loaded from a file

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/test/model/model.jl#L418-L420

the mapping currently used from a point `(x, y)` to index pair is `(i, j) = ((floor(Int, x / dx), floor(Int, y / dy))`. The `round` versus `floor` difference is more of a design choice as to whether to use the nearest grid point for off-grid locations or 'origin' grid point for the cell containing the off-grid location. However, the omission of the `+1` increment is I think a bug as this can potentially lead to zero-indices being produced (which would lead to indexing errors) and is inconsistent with the convention used for computing the grid point coordinates in `get_axes`. 

Similarly when writing the station locations to a file the following conversion is done from indices to points

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/test/model/model.jl#L604-L609

corresponding to `(x, y) = (i  * dx, j  * dx)` rather than `(x, y) = ((i - 1) * dx, (j - 1)  * dx)`. There is a similar off-by-one bug in the following lines

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/test/model/llw2d.jl#L233-L234

when computing the initial height field.

### LINKED ISSUE #230: Optimal proposal particle weights are being computed incorrectly
Currently the optimal proposal computes the particles weights using this function

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/src/ParticleDA.jl#L212-L220

The `observations_mean` argument here [should be the mean of the observed vector](https://link.springer.com/article/10.1023/A:1008935410038) `Y = H * X + V`, `X = F(x) + U`, `U ~ Normal(0, Q)`, `V ~ Normal(0, R)` _given the previous state `x`_ that is `𝔼[Y | x] = H * F(x)`. Currently instead `observation_mean` is set to `HX*`  where `X*` is the state sampled from the optimal proposal.

Currently the previous particle (state) values are overwritten before the log weight computation. While we technically only need access to `H * F(x)` (i.e. the observation operator applied to the forward propagation of the previous state `x` by `F`), for general proposal the weights may depend arbitrarily on both the current and previous particle values so there may be value in having the full values of the previous particles available.

### LINKED ISSUE #231: Summary statistics computed on every time step when not outputted
Currently the (per state index) mean and variance summary statistics are computed on every time step in 

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/src/ParticleDA.jl#L504

These values are however only used (by being written out to a snapshot file) if `filter_params.verbose == true`

https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/src/ParticleDA.jl#L506-L513

with otherwise the values in `filter_data.statistics` just overwritten in the next time step.

Given that computing the summary statistics requires communication across MPI ranks it would make sense to only compute these if they are needed by putting the `get_mean_and_var!` call in a `if filter_params.verbose` conditional block.

matt-graham:
This PR proposes quite a major refactor with two main aims:

  * Assuming all model specific methods expect the model states to be flat vectors. This removes the current assumption of each model state being a three dimensional array with first two axes corresponding to spatial dimensions and the last the state variable, which limits us to models with states defined over a regular rectilinear grid on 2D domains and with the same discretisation for all state variables. With the new abstracted interface we can work with a much broader class of models including non-spatial models and spatial models defined on unstructured meshes. Working with flat vector states should also simplify implementing alternative filtering approaches such as Kalman filtering for linear Gaussian models to use for testing purposes.
  * Decoupling the state particles and observation from the model. Currently the states and observations are assumed to be associated with the `model_data` structure and the 'observations' are simulated using the model in the filtering loop (see #224). This for example makes it difficult to keep independent copies of the states (as potentially required for example to address #230) and also means that it is not possible to use either real observed data or observations simulated from a different model. This PR separates both the state particles and the observations at each time step from the `model_data` struct. As the state particles are now each flat vectors these can simply be maintained in a matrix with each column a state. An option to pass an array of observations to `run_particle_filter` as an optional argument has been added, with the current behaviour of the observations being generated from the model if not specified being maintained as the default if the observations are not specified.

Alongside this the interface that models are expected to implement has been cleaned up significantly, with a more clearly defined minimal set of functions required for just using with the bootstrap particle filter and all use of `Threads.@threads` factored out to the main `ParticleDA` library rather than requiring implementing in each model separately.

@DanGiles: this potentially subsumes some of the functionality added in #226 though with a more major change in interface (and so heavier requirement on updating downstream code such as the SPEEDY integration) so it could still be worth separately merging that in depending on how long we think it will take to get this in place.

The unit tests still need further updates so keeping this as a draft for now.

Fixes #224
Fixes #229 
Fixes #230
Fixes #231


test/model/model.jl
```diff
@@ -297,248 +296,391 @@ function sample_gaussian_random_field!(field::AbstractMatrix{T},
...
-    background_grf::F
-    model_matrices::G
-    rng::H
+function get_station_grid_indices(
+    filename::String,
+    dx::T,
+    dy::T,
+) where T
+    coords = readdlm(filename, ',', Float64, '\n'; comments=true, comment_char='#')
+    return floor(Int, coords ./ [dx dy]) .+ 1
```
DanGiles, line 381: This should be `floor.(Int, coords ./ [dx dy]) .+ 1`

matt-graham: Thanks, just fixed

src/ParticleDA.jl
```diff
@@ -336,10 +336,10 @@ function get_covariance_observation_state_given_previous_state(model_data)
     cov = Matrix{get_state_eltype(model_data)}(
         undef, length(state_indices), observation_dimension
     )
-    for i in state_indices
+    for i in length(state_indices)
```
matt-graham, line 339: This doesn't look correct as the loop is now over an integer (the length of `state_indices`) rather than a collection. I think something like
```Julia
for (i, state_index) in enumerate(state_indices)
```
is probably more what we want here and use `i` to index in to the `cov` array and pass `state_index` to the `get_covariance_state_observation_given_previous_state` function?

src/ParticleDA.jl
```diff
@@ -78,7 +78,7 @@ the type of `model_data`.
 function sample_initial_state! end
 
 """
-    ParticleDA.update_state_deterministic!(state, model_data)
+    ParticleDA.update_state_deterministic!(state, model_data, ip)
```
matt-graham, line 81: What is the specific need for having access to the particle index in `update_state_deterministic!` in SPEEDY (note that the particle index will in general not be the same as `threadid`)? For the particle filter to be correct the state updates shouldn't depend on the particle index. If this is needed to access a thread specific buffer or similar then you can use `threadid` within `update_state_deterministic!`

src/ParticleDA.jl
DanGiles: Yes the enumerate will work, will update the PR now. We were getting a bounds error when solely observing a variable that isn't the first in the flattened array.

src/ParticleDA.jl
DanGiles: Thank you for the suggestion, the threadid is a much better solution. The threadid is needed in the `update_state_deterministic!` in Speedy as individual particles have separate sub folders with the naming structure of rank_id/threadid.

matt-graham:
Marking this ready to review as tests all now update to new interface, but it looks like there is a bug when running with multiple threads 😢. The benchmarks also haven't been updated yet.

src/ParticleDA.jl
```diff
@@ -430,89 +180,127 @@ function run_particle_filter(init, filter_params::FilterParameters, model_params
...
+        end
+    end
+    
+    # Gather final states from all MPI ranks to master
+    @timeit_debug timer "Gather states" flat_global_states = MPI.Gather(
+        vec(states), filter_params.master_rank, MPI.COMM_WORLD
+    )
+    if my_rank == filter_params.master_rank
+        states = reshape(flat_global_states, (size(states, 1), filter_params.nprt))
+    end
```
matt-graham, line 303: This currently gathers the state particles from all ranks to the master rank, which in retrospect is probably a bad idea to have as the default behaviour given the high memory overhead when the number of ranks is high as Tuomas pointed out today (given this is done only once the communication overhead should be relatively minor). I made this change partly as having access to the full final set of states is useful from an analysis perspective and to allow potentially continuing filtering, and also so that the summary statistics correspond to the statistics of these returned particles for testing, but I guess it's probably better to allow the user to manually gather the states on the master rank if needed. Posting this as a reminder for me to fix 😁.

benchmark/benchmarks.jl
```diff
@@ -15,53 +16,173 @@ const SUITE = BenchmarkGroup()
...
+SUITE["Model interface"]["simulate_observations_from_model"] = @benchmarkable (
+    ParticleDA.simulate_observations_from_model($(model), $(n_time_step))
+)
+
+for filter_type in (BootstrapFilter, OptimalFilter), statistics_type in (
+    ParticleDA.NaiveMeanSummaryStat, 
+    ParticleDA.NaiveMeanAndVarSummaryStat,
+    ParticleDA.MeanSummaryStat,
+    ParticleDA.MeanAndVarSummaryStat
+)
```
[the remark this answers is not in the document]
giordano: I don't think this is problematic, [`@benchmarkable`](https://juliaci.github.io/BenchmarkTools.jl/stable/reference/#BenchmarkTools.@benchmarkable-Tuple) doesn't actually run the benchmark, it only defines it, and if doing it in a for loop is more convenient, go for it!

Side note: you're missing the newline at the end of this and many other files :wink:

docs/src/index.md
```diff
@@ -153,42 +178,43 @@ should be set to the total number of observation stations.
...
-   station_distance_x : Distance between stations in the x direction [m]
-   station_distance_y : Distance between stations in the y direction [m]
-   station_boundary_x : Distance between bottom left edge of box and first station in the x direction [m]
-   station_boundary_y : Distance between bottom left edge of box and first station in the y direction [m]
-   ```
-
+2. Provide parameters for an equispaced rectilinear grid of observation stations. The values of these parameters should then be set:
+
+   * `n_stations_x` : Number of observation stations in the x direction
+   * `n_stations_y` : Number of observation stations in the x direction
```
DanGiles, line 184:
```suggestion
   * `n_stations_y` : Number of observation stations in the y direction
```

docs/src/index.md
```diff
@@ -153,42 +178,43 @@ should be set to the total number of observation stations.
...
    As an example, one could set
 
    ```julia
-   nobs : 4
-   station_distance_x : 1.0e3
-   station_distance_y : 1.0e3
-   station_boundary_x : 10.0e3
-   station_boundary_y : 10.0e3
+   n_stations_x=2,
+   n_stations_x=4,
```
DanGiles, line 194:
```suggestion
   n_stations_y=2,
```

docs/src/index.md
```diff
@@ -99,33 +118,40 @@ The particle filter parameters are saved in the following data structure:
...
+
+# Simulate observations from the model to use 
+simulate_observations_from_model(LLW2d.init, input_file, observation_file)
 
-# Run the particle filter using the `init` file defined in the `Model` module.
-run_particle_filter(Model.init, input_file, BootstrapFilter())
+# Run the (optimal proposal) particle filter with simulated observations
+final_states, final_statistics = run_particle_filter(
+  LLW2d.init, input_file, observation_file, OptimalFilter
+)
```
DanGiles, line 154: Should state that if one is running on non-Intel chip you should select ` ParticleDA.NaiveMeanSummaryStat`. Additionally add a subsection stating this. Please see the suggested text.

docs/src/index.md
```diff
@@ -99,33 +118,40 @@ The particle filter parameters are saved in the following data structure:
 ParticleDA.FilterParameters
 ```
```
DanGiles, line 120:
```suggestion
### Output statistics

There are four different output statistics options available:

* NaiveMeanSummaryStat
* NaiveMeanAndVarSummaryStat
* MeanSummaryStat
* MeanAndVarSummaryStat

Information on the differences in methods of calculation can be found in [statistics.jl](https://github.com/matt-graham/ParticleDA.jl/blob/mmg/flat-states/src/statistics.jl)

Please be advised that the MeanSummaryStat and MeanAndVarSummaryStat calculations relie on custom MPI reductions, which inhibits their use on non-Intel architectures. 
In this case please revert to using the Naive calulcations.
```

docs/src/index.md
matt-graham: Now added comment to indicate this.

DanGiles (approved):
Nice work @matt-graham! Serious overhaul, happy for this to be merged.

docs/src/index.md
matt-graham: I've [added some docstrings to the summary stat types](https://github.com/Team-RADDISH/ParticleDA.jl/blob/97f789acb9ab7fdae2534132b41636dce3d10b9a/src/statistics.jl) and added these docstrings plus a short explanation to [the _Usage_ section of the documents index page](https://github.com/Team-RADDISH/ParticleDA.jl/blob/97f789acb9ab7fdae2534132b41636dce3d10b9a/docs/src/index.md#usage).

PR #233, opened 2023-01-27
Tk/thread pinning

tkoskela:
I've added thread pinning opimizations (thanks to @giordano) that could improve performance on Archer2. I would like to test the performance with
- 8 ranks per node, 16 threads per rank
- 16 ranks per node, 8 threads per rank
- 32 ranks per node, 4 threads per rank


extra/weak_scaling/run_particleda.jl
```diff
@@ -1,10 +1,29 @@
...
+# contiguous IDs. This will ensure that
+#  - When running 16 or less threads per rank, all threads will be pinned to the same
+#    NUMA region as their master (sharing a memory controller within Infinity fabric)
+#  - When running 8 or less threads per rank, all threads will be pinned to the same
+#    Core Complex Die
+#  - When running 4 or less threads per rank, all threads will be pinned to the same
+#    Core Complex (sharing a L3 cache)
+
+my_numa, my_id_in_numa = divrem(my_rank, ranks_per_numa) .+ (1, 0)
+pinthreads( numa( my_numa, 1:Threads.nthreads() ) .+ threads_per_rank .* my_id_in_numa )
```
giordano, line 26: With ThreadPinning v0.7.3 you can use simply [`pinthreads(:affinitymask)`](https://carstenbauer.github.io/ThreadPinning.jl/stable/examples/ex_mpi/#Hybrid:-MPI-Threads)

matt-graham:
I think the failed CI jobs on nightly build here may have been due to the same upstream problems that were causing issues in https://github.com/Team-RADDISH/ParticleDA.jl/pull/236#issuecomment-1440353877 but we've now exceeded the 30 day window for being able to re-run workflows. 

This otherwise looks good to merge to me other than @giordano's suggestion above to use `pinthreads(:affinitymask)`. Not sure if there is any other way to trigger a new workflow run than pushing a new commit to this branch as we don't have a `workflow_dispatch` trigger specified to allow running manually. 

Also just noticed this is set to merge in to [`scaling_archer2`](https://github.com/Team-RADDISH/ParticleDA.jl/tree/scaling_archer2) branch - as the `extra/weak_scaling/run_particleda.jl` script currently on `master` hasn't been updated for the changes in #232 and so doesn't run, it would probably be worth merging both those changes and those here in to `master`?

tkoskela:
I don't think this is relevant anymore. If we wanted to look into this it would be worth re-implementing this with the latest version of ThreadPinning.jl.

PR #234, opened 2023-01-27
Run MPI tests with `Base.julia_cmd` instead of `Base.julia_exename`

giordano:
This ensures the same optimisation flags are passed to the processes we spawn, which affect cache invalidation.

I noticed tests with Julia nightly in #233 were hanging.  I believe the problem is the same as https://github.com/JuliaParallel/MPI.jl/pull/706, which took me a while to debug, but at least I did it once and now I know what's the culprit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
           - x64
         threads: [1, 2]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


giordano:
Yes, jobs with Julia nightly are now successful :slightly_smiling_face:  I reported the bug with broken precompilation of Sundials.jl at https://github.com/SciML/Sundials.jl/issues/386.

matt-graham (approved):
Looks good to merge to me.

PR #235, opened 2023-01-27
Upgrade to `MPI.jl` 0.20

giordano:


src/utils.jl
```diff
@@ -85,7 +85,7 @@ function copy_states!(
     end
 
     # Wait for all comms to complete
-    MPI.Waitall!(reqs)
+    MPI.Waitall(reqs, MPI.Status)
```
giordano, line 88: @simonbyrne Tests pass for me locally without deprecation warnings, so I take this is correct, but I was wondering if there's a more efficient way to handle these requests now?

simonbyrne: Unless you need the status objects, you can just do
```suggestion
    MPI.Waitall(reqs)
```

If you want to further improve efficiency, replace `reqs` with a `MultiRequest` or `UnsafeMultiRequest` object.

giordano: Ok, I'll stick to `MPI.Waitall(reqs)` for the time being (I used `MPI.Waitall(reqs, MPI.Status)` because that's what the deprecation message told me to do), we'll perhaps look into using `MultiRequest` later because at the moment we don't work out how many requests we have.  Thanks!

matt-graham (approved):
Looks good to merge to me.

PR #236, opened 2023-01-27
Remove useless dependency on `DifferentialEquations`

giordano:
`OrdinaryDiffEq` is enough, and it should save installation of a couple of bigger packages.

***Edit***: probably needs #235, locally I tested on top of that one.


giordano:
Tests are now successful with most versions of Julia, but failing on nightly (but the problem seems to be upstream in the SciML ecosystem)

matt-graham (approved):
Forgot that I hadn't review this! Thanks for changing @giordano - makes sense to avoid the unnecessary dependencies and changes all look good to me. Tests now passing on nightly as well as it looks like upstream issue with `LinearSolve.jl` has been resolved.

PR #237, opened 2023-02-22
Add linear Gaussian model type + validation notebook

matt-graham:
Adds wrapper type for linear Gaussian models in which filtering can be performed using both Kalman filter (exact) and particle filters (approximate), for the purposes of validating the particle filter implementations. A Pluto notebook validating that the particle filter estimates of the mean and variance parameters of the filtering distributions asymptotically converge to their true values as the ensemble size tends to infinity is also added.

### DIFF
--- a/extra/linear_gaussian_validation.jl
+++ b/extra/linear_gaussian_validation.jl
@@ -8,14 +8,14 @@ using InteractiveUtils
 begin
     import Pkg
     Pkg.activate("../Project.toml")
-	using ParticleDA
-	using LinearAlgebra
-	using PDMats
-	using FillArrays
-	using Random
-	using HDF5
-	using Plots
-	using Statistics
+    using ParticleDA
+    using LinearAlgebra
+    using PDMats
+    using FillArrays
+    using Random
+    using HDF5
+    using Plots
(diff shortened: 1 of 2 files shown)


DanGiles (commented):
Quickly tried running the Pluto notebook. The updated [MPI.Waitall](https://github.com/Team-RADDISH/ParticleDA.jl/blob/9b39f66a0ef3a38b3b3f6b4085b8a7bfbd618010/src/utils.jl#L88)(reqs) is throwing an error.
```
UndefVarError: Waitall not defined
```
Reverting back to MPI.Waitall!() fixes it. So an issue with the previous merge #235, any ideas @giordano?

giordano:
What version of `MPI.jl` do you have in the active environment?
```
]st -m MPI
```
After #235 you'll need to do
```
]resolve
up
```
in the Pkg REPL to re-resolve and update the environmemt.

DanGiles:
Yes, that would do it. Thanks Mose. My misunderstanding of how the Project.toml files work. Happy to merge this now @matt-graham, thanks for the updates.

DanGiles (approved):

PR #252, opened 2023-12-01
Fix Lorenz 63 model filtering bug

matt-graham:
Something seems to have changed in recent `FillArrays` / `Distributions` versions in the behaviour when setting mean and (co)variance parameters for `MvNormal` using `FillArrays` lazy array types `Ones` and `Zeros` rather than explicit array initializers `ones` and `zeros`, which appears to be leading to unintended interactions between particle updates when filtering. This PR switches to using the explicit array initializers and also initializes integrator state specifically to zero rather than `undef` to avoid numerical warnings from `OrdinaryDiffEq`.

### DIFF
--- a/test/models/lorenz63.jl
+++ b/test/models/lorenz63.jl
@@ -2,7 +2,6 @@ module Lorenz63
 
 using Base.Threads
 using Distributions
-using FillArrays
 using HDF5
 using Random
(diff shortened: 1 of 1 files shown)


matt-graham:
Closing as should now be fixed in #256

PR #256, opened 2024-03-12
Remove explicit dependency on `SciMLBase`

giordano:
I believe the reason why we're getting failures in the test with Julia v1.10 in #254 is because the combination of packages `OrdinaryDiffEq@6.58.2` + `SciMLBase@1.97.0` is allowed by the registry but in practice incompatible.  While I believe this should be fixed in the registry by not to allow getting incompatible versions, we should be able to work around the issue by just not requiring `SciMLBase` explicitly and allow the registry to pull more up-to-date versions, with fewer constraints.

Haven't tested locally, but I'm moderately confident this should do the trick.


test/models/lorenz63.jl
```diff
@@ -27,9 +26,9 @@ function get_params(
     return P(; (; (Symbol(k) => v for (k, v) in model_params_dict)...)...)
 end
 
-struct Lorenz63Model{S <: Real, T <: Real}
+struct Lorenz63Model{S <: Real, T <: Real, I}
     parameters::Lorenz63ModelParameters{S, T}
-    integrators::Vector{<:SciMLBase.AbstractODEIntegrator}
+    integrators::Vector{I}
```
matt-graham, line 31: Just to check I'm understanding correctly - this is to avoid the issue described by https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type?

matt-graham: And I guess avoid the direct dependency on `SciMLBase`

giordano: Main point was avoiding loading `SciMLBase` directly, but at the same time adding `I` as extra parameter [avoids the abstract container as field](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-containers), yes.

matt-graham (approved):
This all looks good to me and 🤞 it fixes the issue

matt-graham:
With regards to `FillArrays` - haven't tested recently, but previously I found that something in how I was using `FillArrays` with `Distributions` was causing a bug in the Lorenz 63 model implementation whereby the updates across particles were ending up coupled unintentionally so #252 removes the dependency on `FillArrays` in favour of initializing explicitly. So could potentially merge those changes here and remove `FillArrays` dependency altogether?

PR #259, opened 2024-03-12
Reduce LLW2d model mesh dimensions + number of samples for test

### LINKED ISSUE #257: Adjust number of samples and/or mesh dimension for optimal proposal unit tests with llw2d model
The test set in

https://github.com/Team-RADDISH/ParticleDA.jl/blob/ee916abd0ec50fa582b599b4072514e9bef237d5/test/runtests.jl#L515-L533

with linear long wave 2D model is taking ~7 minutes to run in CI jobs. We should either reduce maximum number of samples we check for convergence over for that model or reduce mesh dimension from current default of 41 by 41 (which gives us a ~1600 times 1600 covariance matrix to check).

matt-graham:
Resolves #257 

Reduces dimensions of mesh used for `LLW2d` model in optimal proposal model interface unit tests for `(41, 41)` to `(11, 11)` so that we are checking 121×121 covariance matrices rather than 1681×1681.


giordano (approved):

matt-graham:
With both the reduced mesh dimension and number of samples down to ~40 seconds on Actions runners, though not sure why it's still so slow as it's running in around 4 seconds locally for me.

giordano (approved):

PR #263, opened 2024-03-13
Ignore result of runs with nightly builds in CI job status

matt-graham:
This replicates the approach in https://github.com/JuliaTime/TimeZones.jl/pull/418 also discussed in https://github.com/actions/runner/issues/2347 and https://github.com/orgs/community/discussions/15452, for trying to have failures of the CI runs on Julia nightly version not cause the overall workflow status to be a failure. I _think_ that this would mean the workflow status (and so status badge, tick / cross indicator in commit history) would remain green on nightly failures while still showing the failures in the individual job statuses, though I'm not totally sure on this so partly opening up this PR to test. Ideally we would have some alternative visual indicator for the overall workflow status to highlight some optional jobs failing, but given the above linked discussions suggest this has been a long requested GitHub Actions feature with no movement as of yet, I would say having the workflow statuses indicating passing even when nightly builds are failing is overall preferable to having them always showing as failing in this case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
     env:
       JULIA_NUM_THREADS: ${{ matrix.threads }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.version == 'nightly' }}
     strategy:
       fail-fast: false
       matrix:


matt-graham:
Looks like this doesn't make any difference to the status indicator on the PR but the status badge does appear to be showing as passing:

```Markdown
![Build Status](https://github.com/Team-RADDISH/ParticleDA.jl/actions/workflows/ci.yml/badge.svg?branch=mmg/continue-on-error-nightly)
```

Build Status

giordano (approved):
This also appears green in the [actions page](https://github.com/Team-RADDISH/ParticleDA.jl/actions/workflows/ci.yml?query=branch%3Ammg%2Fcontinue-on-error-nightly).  I still think the old UI of Travis was nicer, but I guess we have to do with what we have now.  Thanks for this!

PR #264, opened 2024-03-13
Add extra links to README and improve description

matt-graham:
Adds badges to README for latest stable version of documentation and GMD journal article (will need to be updated once the accepted version is available online). Also tweaks the description a little bit to mention thread-based parallelism and add a link to ARC website. I also noticed there was a `unit_test.h5` HDF5 file we don't appear to use anywhere that was added in #218, I assumne accidentally, so have removed this.


README.md
```diff
@@ -1,12 +1,18 @@
 # ParticleDA
 
+[![Stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://team-raddish.github.io/ParticleDA.jl/stable/)
 [![Dev](https://img.shields.io/badge/docs-dev-blue.svg)](https://team-raddish.github.io/ParticleDA.jl/dev/)
 [![Build Status](https://github.com/Team-RADDISH/ParticleDA.jl/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/Team-RADDISH/ParticleDA.jl/actions/workflows/ci.yml?query=branch%3Amaster)
 [![Coverage](https://codecov.io/gh/Team-RADDISH/ParticleDA.jl/branch/main/graph/badge.svg)](https://codecov.io/gh/Team-RADDISH/ParticleDA.jl)
```
giordano, line 6: Unrelated to this PR, I just realised this is wrong because our default branch is called `master`, not `main`.  I'd be happy with renaming it to `main`, but I guess we should be consistent in these badges.  For the rest, this looks good

unit_test.h5
giordano, line 1: Should we gitignore all `*.h5` files?  If we need to include them for testing we can always `git add --force` them.

README.md
matt-graham: I'm also happy with renaming `master` &rightarrow; `main`. Looking for `master` references in code as well as updating README we'd need to update Actions workflows. We also have lots of usages of `master_rank` which I guess it would sort of make sense to also update to `main_rank` or similar (though technically this would be a breaking change as we expose `master_rank` as a filter parameter). For purposes of this PR I guess probably easiest to update Codecov badge to point to `master` and do a separate PR for updating references to `main` (+ rename branch)?

unit_test.h5
matt-graham: Yeah that seems sensible - will add a rule to `.gitignore`.

README.md
giordano, line 1: I now renamed the branch, so should replace `master` -> `main` everywhere.  Sorry! :sweat_smile:

giordano: Potentially yes, but it can be done in a separate PR.  Also, it'd probably be a technically breaking change, as I think we expose the name to users at https://github.com/Team-RADDISH/ParticleDA.jl/blob/16180555362f53669d6bde0a18622fdf6e7aed4c/src/params.jl#L6

matt-graham:
Looks like there is some implicit reference to `master` branch in benchmarks workflow from the fact it fails with
```
fatal: couldn't find remote ref refs/heads/master
```
when running
```Julia
using PkgBenchmark, BenchmarkCI

BenchmarkCI.judge(
    PkgBenchmark.BenchmarkConfig(
        env = Dict(
            "JULIA_NUM_THREADS" => "2",
        ),
    ),
)
```
but not sure how to fix - @giordano any ideas?

giordano (approved):

PR #260, opened 2024-03-12
[CI] Add dependabot, use newer workflows, use concurrency groups

giordano:

PR #266, opened 2024-03-13
Run different CI jobs only when relevant files are touched

giordano:
This will reduce CI workload by not running unnecessary jobs.


matt-graham (approved):
Tagging https://github.com/UCL-ARC/green-team/issues/76 in case we need to demonstrate we followed through on the pledge to reduce unnecessary CI usage 😁

giordano:
Also #260 was relevant with the use of the concurrency groups :grin:  Oh, I just added also the timeouts. I wish doing all of this was simpler, at least default timeout could be much less than 6 hours!

PR #287, opened 2024-11-15
Register custom MPI operator for aarch64

giordano:
I still need to test this, but opening to remind myself to look into this.


giordano:
Well, now we're testing this on CI on aarch64-darwin, and it seems to be working fine :slightly_smiling_face: CC: @sguillas.

giordano:
For the record, I ran the tests successfully on both A64FX and Nvidia Grace (besides the fact we're now testing the package on aarch64-darwin, too)

PR #288, opened 2024-11-15
Update benchmarks environment to Julia v1.10

giordano:


giordano:
Just moving to v1.10 shaved off 12+ minutes from CI time: https://github.com/Team-RADDISH/ParticleDA.jl/actions/workflows/benchmark.yml?query=is%3Asuccess

PR #294, opened 2024-12-17
Move functions for testing models into main package

matt-graham:
Currently we have some general purpose functions for testing that models adhere with the expected interface (define the required methods) in `tests/runtests.jl`, which we use to check the test models in `tests/models`. These include a function to check model implements interface required for general filters correctly, the additional methods required for using the locally optimal proposal filter and a function to validate filters when applied to linear-Gaussian models against the Kalman filter.

To simplify testing models defined by users or in other repositories such as `ParticleDA-UseCases` it would be useful for these functions to be available within the main `ParticleDA` package which is what this PR does.


giordano (approved):
So, I was initially a bit hesitant to add `Test` as a dependency.  There are a couple of packages which can help with enforcing interfaces (until the day [this will be a built-in feature](https://hackmd.io/BbEw0_B4Q8uDSS34LOvpCw)):

* [`RequiredInterfaces.jl`](https://github.com/Seelengrab/RequiredInterfaces.jl)
* [`Interfaces.jl`](https://github.com/rafaqz/Interfaces.jl)

but they both internally depend on `Test` for the same reason :sweat_smile: Not sure it's worth looking into those packages right now, since the work for testing the interfaces was already there, and this PR is UP already.

matt-graham:
> So, I was initially a bit hesitant to add `Test` as a dependency. There are a couple of packages which can help with enforcing interfaces (until the day [this will be a built-in feature](https://hackmd.io/BbEw0_B4Q8uDSS34LOvpCw)):
> 
>     * [`RequiredInterfaces.jl`](https://github.com/Seelengrab/RequiredInterfaces.jl)
> 
>     * [`Interfaces.jl`](https://github.com/rafaqz/Interfaces.jl)
> 
> 
> but they both internally depend on `Test` for the same reason 😅 Not sure it's worth looking into those packages right now, since the work for testing the interfaces was already there, and this PR is UP already.

Thanks for the pointers, I had wondered if there was existing packages for doing helping do these sort of checks! Agree also that depending on `Test` in package feels a bit weird but also couldn't see an easy way to avoid while still getting relatively useful feedback on test failures. 

I think a pre-requisite of using those packages would be to make our interface more well defined by having some abstract type models are expected to derive from (and also using this to differentiate between general models and those with the conditional-Gaussian structure that allows using optimal proposal), which I think is something we have discussed at various points. I think that would be a worthwhile change to do but probably better as a separate PR. I'll create an issue to remind us to look at this and also the above packages at some point!

